Issue #24, opened 2020-02-05
Packages and feedstocks we should test this


ocefpaf:
Pretty much any Python package in https://github.com/regro/cf-graph-countyfair/blob/master/status/could_use_help.json, bonus point if this creates a recipe that passes the CI!

- `hypothesis` (almost daily updates and high change of changing metadata)
- `boto`/`botocore` (one is tied to another and with the proper metadata `boto`'s PR should fail until `botocore` is merged)
- `gsw` (compiler)
- `matplotlib` (messy outputs)
- `pytest` (lots of new syntax for package metadata

marcelotrevisani:
At least, for now, I generate recipes for these packages:

- ``hipothesis``: https://gist.github.com/marcelotrevisani/31c45c94ed2305056cdd0caa508cce79
- ``boto``: https://gist.github.com/marcelotrevisani/c5385a33da4e966de48542040e73845e
- ``botocore``: https://gist.github.com/marcelotrevisani/a57a09007128923e549b8b8c758f7516
- ``gsw``: https://gist.github.com/marcelotrevisani/52c87047f7af20fe78f1b854d0a386de
- ``matplotlib``: https://gist.github.com/marcelotrevisani/aa0c83dad03829ef3ffc9ae4e78cb263
- ``pytest``: https://gist.github.com/marcelotrevisani/fc4aff4de383703f23f74dbed1b39f9b
- ``requests``: https://gist.github.com/marcelotrevisani/28f63677009685eb865449ece713f1cd


For now, the project is supporting just the metadata which comes from pypi, but I will start the development to do the injection in the distutils and with that I will be able to recover some information when the pypis` metadata is missing.

Those projects which they provided well the pypi metadata (such as requests and pytest) the recipe generated is pretty close to the current on ``conda-forge`` 

From time to time I will try to update those gist when I changed the code

marcelotrevisani:
After the integration of #32 it will improve those recipes generated

ocefpaf:
The `gsw` recipe needs a compiler and the conda-style numpy pin. We don't need to add support for that right now but we should open issues to track the development of those features.

As far as I know there is no metadata to declare that yet, only ideas. If we will try to be smart we should look for numpy header include in the `setup.py`, and the use of the `Extension` class for compilers.

marcelotrevisani:
[...]

PR #32, opened 2020-02-11
Add methods to do the injection in the distutils

marcelotrevisani:
Related to #32 and #24


marcelotrevisani:
It still needs to create some functions to mege the pypi metadata and sdist metadata

marcelotrevisani:
tests are failing but they are not related to this PR
For some reason ``coverage`` stopped to work. it needs investigation, I will take a look on that later

PR #36, opened 2020-02-14
Possibility to call grayskull to generate multiple recipes

marcelotrevisani:


marcelotrevisani:
So it will be possible to generate several recipes from one command
```
grayskull pypi pytest gsw matplotlib
```

each recipe will be in a different folder

PR #42, opened 2020-02-17
Add support for fortran

marcelotrevisani:


marcelotrevisani:
This error is occuring just on ``linux`` and in the CI.
I created a new environment from scratch and the tests are passing...
It needs investigation on azure

PR #132, opened 2020-05-25
Return a better message when not executing grayskull correctly

marcelotrevisani:


grayskull/__main__.py
```diff
@@ -18,8 +18,8 @@
 
 
 def main(args=None):
-    if args is None:
-        args = sys.argv[1:]
+    if not args:
+        args = sys.argv[1:] if sys.argv[1:] else ["--help"]
```
marcelotrevisani, line 22:
```suggestion
    if args:
        return
```

marcelotrevisani, line 22:
```suggestion
    if args:
        raise Exception()
```

PR #169, opened 2020-08-12
Add docker

mariusvniekerk:
Don't know if we want to use something like precommit to ensure that this is sorted automatically for user-ease


marcelotrevisani:
> Don't know if we want to use something like precommit to ensure that this is sorted automatically for user-ease

I configure pre-commit to check yaml files, but it is not sorting the keys. But it would be nice to sort it, indeed!

marcelotrevisani:
Thanks for adding it! :)

PR #179, opened 2020-08-16
_fix_list_requirements: ensure_list

nehaljwani:


grayskull/utils.py
```diff
@@ -64,3 +65,12 @@ def get_local_modules(sdist_folder: str) -> List:
...
     return result
+
+
+def ensure_list(arg) -> List:
+    if isinstance(arg, types.StringTypes) or not hasattr(arg, "__iter__"):
+        if arg is not None:
+            arg = [arg]
+        else:
+            arg = []
+    return arg
```
marcelotrevisani, line 76: I think we can try to simplify 
```suggestion
    """Convert object to a list
    
    :param arg: Python object which will be converted to a list
    :return: Object converted
    """
    if arg is None:
        return []
    if not isinstance(arg, list):
        return [arg]
    return arg
```
What do you think?

and could you please just add a unit test to cover this help function would be nice :)

nehaljwani:
I don't know what's up with the failing test. Passes locally and seems unrelated to the change.

marcelotrevisani:
> I don't know what's up with the failing test. Passes locally and seems unrelated to the change.

Yes indeed, it seems to be a flake error. Thanks for fixing this bug!

Issue #184, opened 2020-08-29
Import fails using Python 3.8.5


dmyersturnbull:
**Describe the bug**

**To Reproduce**
1. Install Miniconda on Windows 10 as of 2020-08-28.
2. Run:
   ```bash
    conda create --name fresh \
   --override-channels --channel conda-forge \
    --force --yes python=3.8
    pip install tox poetry grayskull
    ```
2. Run `conda activate fresh && grayskull --version`

**Outputs**

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\ProgramData\Miniconda3\lib\site-packages\grayskull\base\factory.py", line 4, in <module>
    from grayskull.pypi import PyPi
  File "C:\ProgramData\Miniconda3\lib\site-packages\grayskull\pypi\__init__.py", line 1, in <module>
    from grayskull.pypi.pypi import PyPi  # noqa
  File "C:\ProgramData\Miniconda3\lib\site-packages\grayskull\pypi\pypi.py", line 21, in <module>
    from grayskull.base.base_recipe import AbstractRecipeModel
  File "C:\ProgramData\Miniconda3\lib\site-packages\grayskull\base\base_recipe.py", line 11, in <module>
    from ruamel.yaml import YAML, CommentToken
ImportError: cannot import name 'YAML' from 'ruamel.yaml' (unknown location)
```

**Environment:**
 - conda list:
  ```
# Name                    Version                   Build  Channel
ca-certificates           2020.6.20            hecda079_0    conda-forge
certifi                   2020.6.20        py38h32f6830_0    conda-forge
openssl                   1.1.1g               he774522_1    conda-forge
pip                       20.2.2                     py_0    conda-forge
python                    3.8.5           h60c2a47_7_cpython    conda-forge
python_abi                3.8                      1_cp38    conda-forge
setuptools                49.6.0           py38h32f6830_0    conda-forge
sqlite                    3.33.0               he774522_0    conda-forge
vc                        14.1                 h869be7e_1    conda-forge
vs2015_runtime            14.16.27012          h30e32a0_2    conda-forge
wheel                     0.35.1             pyh9f0ad1d_0    conda-forge
wincertstore              0.2                   py38_1003    conda-forge
  ```
 - pip list: a bit redundant
 - grayskull --version: **fails**

dmyersturnbull:
Also tried:

```bash
pip uninstall grayskull
pip install grayskull=0.7.4
```

Same issue.

[...]

PR #185, opened 2020-08-29
Require ruamel.yaml >=0.16.10

dmyersturnbull:
Fixes issue #184

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
     python_requires=">=3.7",
     install_requires=[
         "requests",
-        "ruamel.yaml >=0.15.3",
+        "ruamel.yaml >=0.16.10",
         "ruamel.yaml.jinja2",
         "stdlib-list",
         "pip",


dmyersturnbull:
On a similar note, it may be worth using more complete version ranges:

```
python_requires=">=3.7, <=3.9"
```

And:
```
        "requests",
        "ruamel.yaml >=0.16, <1.0",
        "ruamel.yaml.jinja2 >=0.2, <1.0",
        "stdlib-list >=0.7, <1.0",
        "pip >=20",
        "setuptools >=30.3",
        "rapidfuzz >=0.7.6, <1.0",
        "progressbar2 >=3, <4",
        "colorama >=0.4, <1.0",
```

marcelotrevisani:
You are completely right, it is better to limit a bit more the dependencies or to pin it. However, I was thinking to let it more loosely and just limit it when necessary, I know it is not the best alternative. I just would like to remove the upper bound constrains and if you like can you add those lower bounds constrains please?

marcelotrevisani:
After it gets merged we can generate a new version for grayskull

marcelotrevisani:
I will merge it as it is for now, we can update the requirements later then
thanks for fixing this

marcelotrevisani (approved):

PR #206, opened 2020-11-26
Add PyPI mappings for setuptools-scm and python-htcondor

duncanmmacleod:
This PR adds PyPI mappings for a couple of packages I've seen with naming quirks lately.

I hope this sort of change proposal is appropriate.

### DIFF
--- a/grayskull/pypi/config.yaml
+++ b/grayskull/pypi/config.yaml
@@ -314,6 +314,10 @@ sagemaker_mxnet_training:
   conda_forge: sagemaker_mxnet_container
   import_name: sagemaker_mxnet_container
 
+setuptools_scm:
+  conda_forge: setuptools-scm
+  import_name: setuptools_scm
+
 sounddevice:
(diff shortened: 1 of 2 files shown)


marcelotrevisani:
``setuptools_scm`` is not really needed, but if you want to standardise it to ``setuptools_scm``, could you please fix the tests?
I believe there are a couple of tests expecting ``setuptools-scm``

duncanmmacleod:
> `setuptools_scm` is not really needed, but if you want to standardise it to `setuptools_scm`, could you please fix the tests?
> I believe there are a couple of tests expecting `setuptools-scm`

The idea here is to standardise it to `setuptools-scm` since that is the actual conda-forge package that contains the modules - `setuptools_scm` is just an empty metapackage, so avoiding using seems like an easy way to avoid one more download.

I'll fix the tests now.

marcelotrevisani (approved):

marcelotrevisani:
Thanks for your contribution! :)

PR #218, opened 2021-03-19
add ability to generate conda recipes for whl packages

madhur-tandon:


marcelotrevisani:
Hi @madhur-tandon ,

Sorry, I didn't see this PR earlier than normal. The modification looks good, are you willing to add some tests to cover the recipe creation from wheels please?

marcelotrevisani:
If I do not answer in a couple of days please feel free to ping me, sometimes some notifications sometimes might get lost in the bunch of notifications I receive

marcelotrevisani:
the bot also found some unused variables and unused imports, could you please take a look?

beenje:
@madhur-tandon are you still interested in moving that forward?
Otherwise I could have a look.

marcelotrevisani:
> @madhur-tandon are you still interested in moving that forward?
> Otherwise I could have a look.

I believe this PR is stale, the author didn't answer anymore :/

It would be awesome to add this feature to ``grayskull``, if you could work on it, it would be really really cool! :)

beenje:
@marcelotrevisani, I had a quick look and from my understanding, a wheel doesn't include any information about `setup_requires`.
I don't see how we could produce a recipe from only a wheel (no information about the host/build requirements).
Or am I missing something?

marcelotrevisani:
> @marcelotrevisani, I had a quick look and from my understanding, a wheel doesn't include any information about `setup_requires`.
> I don't see how we could produce a recipe from only a wheel (no information about the host/build requirements).
> Or am I missing something?



> @marcelotrevisani, I had a quick look and from my understanding, a wheel doesn't include any information about `setup_requires`.
> I don't see how we could produce a recipe from only a wheel (no information about the host/build requirements).
> Or am I missing something?

Yes, you are correct, the wheel does not have information about the ``setup_requires``, but it does have information about the run requirements and entry points

marcelotrevisani:
> @marcelotrevisani, I had a quick look and from my understanding, a wheel doesn't include any information about `setup_requires`.
> I don't see how we could produce a recipe from only a wheel (no information about the host/build requirements).
> Or am I missing something?

that is an old fight that people have with pypi :/
and also how difficult is to recover requirements from python packages...
They could make our life easier, but that is not the case

beenje:
> Yes, you are correct, the wheel does not have information about the `setup_requires`, but it does have information about the run requirements and entry points

Indeed, there are a lot of information we can retrieve, but if some is missing, when would using a wheel be something we want to do? Is it when the sdist doesn't exist?

I like to use `setuptools-scm` for example and this information can't be retrieved from the wheel, meaning that we couldn't produce an accurate conda recipe for such package (without the sdist).

marcelotrevisani:
> > Yes, you are correct, the wheel does not have information about the `setup_requires`, but it does have information about the run requirements and entry points
> 
> 
> 
> Indeed, there are a lot of information we can retrieve, but if some is missing, when would using a wheel be something we want to do? Is it when the sdist doesn't exist?
> 
> 
> 
> I like to use `setuptools-scm` for example and this information can't be retrieved from the wheel, meaning that we couldn't produce an accurate conda recipe for such package (without the sdist).
> 
> 

Indeed, it would be really nice if we can retrieve such information. There are some cases da people does not upload the sdist package to pypi and there is other case that people want to generate the recipe from the wheel directly. I would like to propose one thing, we can try to extract the information from the wheel where the repository resides and with that we can checkout the repository in a specific version ou try to get from the tarball and with that we can run grayskull to also get the bike build and host. If such thing is not possible we just add pip and Python to host and in the run requirements we add all the information that can be found in the wheel. 
If you think that is too much work just the latter would be nice smart already. 
What do you think?

beenje:
I'll try to continue what I started and see where it goes.

agronholm:
What do you need `setup_requires` for? If you have a wheel, that setup phase has already been completed so those requirements are not needed anymore.

marcelotrevisani:
The idea is to still try to get the sdist package or repository if it is available and build it from scratch.

sugatoray:
Hi @marcelotrevisani and @beenje ,

This will be great if making packages from the whl files are possible. Any updates on the PR?

beenje:
Hello,
No sorry I didn't make progress on that. For my use case, creating the recipe from the sdist is working well.
I still would like to finish what I started...

Note that this is to create a conda recipe from a wheel.
For pure python packages, it should be possible to create directly a conda package from a wheel. But this is something completely different. Another project did some tests: https://github.com/takluyver/wheel2conda but it would require quite some work.

marcelotrevisani:
Closing this as it seems to be stale

Issue #228, opened 2021-06-14
[BUG] Grayskull allows recipes with _ when name should have -


maresb:
**Describe the bug**

Running
```bash
grayskull pypi slack_sdk
```
generates a recipe named `slack_sdk`, even though the PyPi package name is `slack-sdk`.

**Expected behavior**

An error message that the name `slack_sdk` doesn't correspond to a PyPi package.

**Outputs**
```text
{% set name = "slack_sdk" %}
{% set version = "3.6.0" %}
...
```

**Environment:**
 - grayskull --version `0.8.5`

**Additional context**

I suspect that `grayskull` is following the PyPi redirect to remove underscores.

This issue led to unintentionally create a duplicate feedstock:

https://github.com/conda-forge/slack-sdk-feedstock/issues/10#issuecomment-860884130

marcelotrevisani:
Hi @maresb ,

I believe the "fix" in this case is to just give an warning to the user that the package already exist on ``conda-forge``.

marcelotrevisani:
cc @ForgottenProgramme

maresb:
@marcelotrevisani or to not follow redirects on PyPi?

marcelotrevisani:
The project name is ``slack_sdk``,
https://github.com/slackapi/python-slack-sdk/blob/d3615007df2b2f7fd9da1d7afbcd88fe873a3590/setup.py#L250

It is mimicking the pypi behaviour as it is generating the recipes looking for pypi
I think grayskull should follow the name of the project, so it should be ``slack_sdk`` then

maresb:
Ah! Very good point. I guess PyPi is blindly converting underscores to dashes?

It seems not so easy to search for duplicates: exponentially many possibilities in the number of underscores plus dashes

I wish there was clear guidance for a standard on conda-forge. I have been searching and asking for a while with no answers.

ForgottenProgramme:
@maresb 
>I guess PyPi is blindly converting underscores to dashes?

Yes.

ForgottenProgramme:
Also, @maresb, when you say this:
>It seems not so easy to search for duplicates: exponentially many possibilities in the number of underscores plus dashes

You mean searching for duplicates on conda-forge, right?
Because if PyPI converts all underscores to hyphens, it's not possible to have duplicates. Isn't it?

maresb:
@ForgottenProgramme, exactly.

[...]

PR #230, opened 2021-06-21
Recipe names follow PyPI naming (only hyphens, no underscores)

ForgottenProgramme:
Added a 'canonicalize_name' method to change all underscores to hyphen in the recipe name. 

user input: ``grayskull pypi slack_sdk``
Recipe generated: ``slack-sdk``

Solves #228 

See this Gitter conversation: [June 18, 2021 2:03 PM](https://gitter.im/conda_grayskull/community?at=60cc5a48d20143617e504c2e)

22/06/2021
Edit:

- We want to decide how recipes generated via grayskull are named.
- Presently grayskull names the recipes (after finding the project on PyPI) same as what is entered by the user. ``grayskull pypi <project_name_entered_by_user>``
- The alternatives: 
1. Use the PyPI name from metadata
2. Use the 'original' project name from metadata


ForgottenProgramme:
slack_sdk

The messages displayed still show 'slack_sdk'. Maybe that needs to be changed as well? 
I'll make a different PR for that.

ForgottenProgramme:
@marcelotrevisani

grayskull/pypi/pypi.py
```diff
@@ -55,9 +55,15 @@ def __init__(
...
         self._setup_metadata = None
         self._is_arch = False
+        name = self._canonicalize_name(name)
         super(PyPi, self).__init__(name=name, version=str(version) if version else None)
         self["build"]["script"] = "<{ PYTHON }} -m pip install . -vv"
 
+    @staticmethod
+    def _canonicalize_name(name:str ):
+        """replace all underscores with hyphens in package name"""
+        return name.replace("_","-")
```
ocefpaf, line 65: While most projects use `-` in the pkg name and `_` in the import name that is not always true. I'm not sure we have enough metadata to figure that out automatically, if so, we should. If not, the implementation here is making an assumption that may, in some rare cases, fail.

TL;DR I'm not against it but I would prefer a metadata based approach first and a fallback to something we all agree on.

marcelotrevisani: I agree with @ocefpaf 

we recover the project name from ``setup.py`` and pypi metadata and after it grayskull try to merge both information, right now it discard the name and does not set it, one way to do it would be:

in the method ``_extract_requirements`` inside of the ``PyPi`` class online [781](https://github.com/conda-incubator/grayskull/blob/2e2ab84beb8dbb71a23eb43abcecb5b48eba949a/grayskull/pypi/pypi.py#L781) we have:
```python
        name = metadata["name"]
```
so, one thing which we can do is:
```python
        name = metadata["name"]
        self.set_jinja_var("name", name)
```

that should solve it for now

marcelotrevisani:
@ocefpaf , this project ``slack_sdk`` has different names on pypi and on project
pypi metadata is showing as ``slack-sdk`` https://pypi.org/pypi/slack-sdk/json
however, the project name is: ``slack_sdk`` https://github.com/slackapi/python-slack-sdk/blob/148fbb181a2ba314518118aaa8e590d17ae3eb10/setup.py#L250

which one should we follow? I think it would be better to follow the project name instead of pypi, because you know how problematic pypi is...

ForgottenProgramme:
@ocefpaf 
Hello, Filipe!
Would you please add your comments on this issue in this grayskull gitter chat? https://gitter.im/conda_grayskull/community
It would be a lot easier to discuss there.

ocefpaf:
We are hitting the issues where PyPI doesn't care about _ vs - but we do because conda isn't python only. What we did in the past, for cases like these, was to use multiple outputs to provide both names. However, that is not an official practice. 

Some people dislike having an empty pkg just to reserve a name.

TL;DR let's pause this and discuss with more core members to see if we can agree on something.

maresb:
I was just looking at some old issues, and I noticed that this was closed. What was the conclusion in the end?

PR #237, opened 2021-08-08
Edited code comment to make it clearer

### LINKED ISSUE #233: Confusing comment in the _injection_disutils method of the PyPI class
See [here](https://github.com/conda-incubator/grayskull/blob/0fc0cc725ba6e8a37b901d8d36cf6603b8cd7abc/grayskull/pypi/pypi.py#L200).

"This method is necessary because some information are missing from the
pypi metadata and also for those packages which the pypi metadata is missing. "

@marcelotrevisani , would you please clarify what it's trying to say, specially in the second line?

ForgottenProgramme:
Fixes #233


marcelotrevisani (approved):

marcelotrevisani:
Thanks for fixing this! :)

PR #238, opened 2021-08-11
[Fix #224] Added check for scripts in metadata and wrote tests

### LINKED ISSUE #224: [BUG] Package cannot be noarch if scripts argument is used
**Describe the bug**

If a package has the `scripts` argument in _setup.py_ it cannot be [`noarch`](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python). Currently, this is not taken into account by Grayskull.

**To Reproduce**
1. Create package

**Expected behavior**

`noarch: python` should not be included in the `build:` section.

**Outputs**
If applicable, add the output to help explain your problem.

One example is [`remove_dagmc_tags`](https://github.com/svalinn/remove_dagmc_tags) which has the following [_setup.py_](https://github.com/svalinn/remove_dagmc_tags/blob/c529c688959647faaebba0a932f96c0c10891f61/setup.py#L17)

```python
setuptools.setup(
    name="remove_dagmc_tags",
    version="0.0.5",
    author="DAGMC development team",
    description="A tool for selectively removing tags such as the graveyard from DAGMC h5m files.",
    long_description=long_description,
    long_description_content_type="text/markdown",
    url="https://github.com/svalinn/remove_dagmc_tags",
    packages=setuptools.find_packages(),
    zip_safe=True,
    package_dir={"remove_dagmc_tags": "remove_dagmc_tags"},
    scripts=['remove_dagmc_tags/remove-dagmc-tags'],
    package_data={
        "remove_dagmc_tags": [
            "requirements.txt",
            "README.md",
            "LICENSE",
        ],
        "tests": [
            "dagmc.h5m"
        ]
    },
    install_requires=[
        "numpy"
    ],
    # pymoab is also required for this package but is not available via pip
    # install
    tests_require=["pytest-cov"],
)
```

The output here is

```yaml
build:
  number: 0
  noarch: python
  script: {{ PYTHON }} -m pip install . -vv
```

**Environment:**
 - https://www.marcelotrevisani.com/grayskull

**Additional context**

ForgottenProgramme:
Fixes #224

### DIFF
--- a/grayskull/cli/stdout.py
+++ b/grayskull/cli/stdout.py
@@ -19,9 +19,7 @@ def print_msg(msg: str):
 def manage_progressbar(*, max_value: int, prefix: str):
     if CLIConfig().stdout:
         with progressbar.ProgressBar(
-            widgets=deepcopy(WIDGET_BAR_DOWNLOAD),
-            max_value=max_value,
-            prefix=prefix,
+            widgets=deepcopy(WIDGET_BAR_DOWNLOAD), max_value=max_value, prefix=prefix,
         ) as bar:
             yield bar
     else:
--- a/grayskull/cli/stdout.py
+++ b/grayskull/cli/stdout.py
@@ -19,7 +19,9 @@ def print_msg(msg: str):
 def manage_progressbar(*, max_value: int, prefix: str):
     if CLIConfig().stdout:
         with progressbar.ProgressBar(
-            widgets=deepcopy(WIDGET_BAR_DOWNLOAD), max_value=max_value, prefix=prefix,
+            widgets=deepcopy(WIDGET_BAR_DOWNLOAD),
+            max_value=max_value,
+            prefix=prefix,
         ) as bar:
             yield bar
     else:
--- a/grayskull/pypi/pypi.py
+++ b/grayskull/pypi/pypi.py
@@ -1000,8 +1000,7 @@ class PyPi(AbstractRecipeModel):
         if not pypi_metadata["requires_python"]:
             return None
         req_python = re.findall(
-            r"([><=!]+)\s*(\d+)(?:\.(\d+))?",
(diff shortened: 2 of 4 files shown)


ForgottenProgramme:
@jaimergp

marcelotrevisani:
just for reference:
https://pre-commit.com/

ForgottenProgramme:
Hey, @marcelotrevisani !
When I run pre-commit for all files, this is what I get:

image

marcelotrevisani:
yeap, the first time will fail because it is showing you the problems, but it will auto fix the black stuffs
if you run it again it should pass

ForgottenProgramme:
@marcelotrevisani
Ahh, got it. Thanks. :))

marcelotrevisani (approved):

ForgottenProgramme:
@marcelotrevisani , 
just one last thing: 
I am printing a warning message that 'scripts were detected. marking the package as non-noarch' in RED font. 
@jaimergp , pointed out that Red font is usually for errors. We could probably use regular white font color instead. But then again, white font color will not stand out from all the other text in white. We really want to the user to know that we have marked their package as non-noarch because it had scripts. 
What do you think?

marcelotrevisani:
> @marcelotrevisani ,
> just one last thing:
> I am printing a warning message that 'scripts were detected. marking the package as non-noarch' in RED font.
> @jaimergp , pointed out that Red font is usually for errors. We could probably use regular white font color instead. But then again, white font color will not stand out from all the other text in white. We really want to the user to know that we have marked their package as non-noarch because it had scripts.
> What do you think?

You can use Yellow in this case or convert that to use
```python
logging.warning("...")
```

marcelotrevisani:
Please let me know when you are done with your modifications then, and I will review and integrate it

marcelotrevisani (approved):

PR #240, opened 2021-09-09
Add annoy to pypi config

candalfigomoro:


marcelotrevisani (approved):

marcelotrevisani:
I think you will need to run pre-commit on the modifications, that should solve the CI issue

PR #244, opened 2021-10-15
Fix #161: Generate recipes from Github 

### LINKED ISSUE #161: Add possibility to generate python recipes from git repositories
Add possibility to generate python recipes from git repositories as well

ForgottenProgramme:
Fixes #161 

**Generate recipes for Github packages**
This PR attempts to add to Grayskull the ability to generate recipes for those Python packages that exist as Github repositories but do not exist on PyPI. 

**Motivation**
- Presently Grayskull generates recipes _only for those Python packages that are available on PyPI_. This pre-requisite leaves out a number of Python packages from being in Grayskull's 'recipe-generatable' list. 
- Grayskull's ability to generate recipes for Github packages can reduce internal packaging efforts in companies.   

**Rationale**
- Grayskull extracts package metadata from two sources to generate the recipe - sdist metadata and pypi metadata. For a package not published on pypi, pypi metadata is obviously not present. The existent Grayskull machinery to extract sdist metadata was tweaked so that it could also be used to extract metadata for Github packages.
- Version of a github package as specified by the user and the original version name might differ (for example: 1.0.0 / v1.0.0 / ver1.0.0/ ), string matching is used and the 'tag' with the closest name is considered for recipe generation.

**Backwards Compatibility**
Backwards Compatibility is maintained. 

**Alternatives Links to prototype or a sample implementation of the proposed change.**
During the implementation of this new feature, subclassing and creating a separate machinery altogether for Github packages was considered. It is still under consideration and might be implemented in future. 

**Sample Implementation**
- We still use the pypi entry point
- We let the name be either a pypi package name or a GitHub repository URL
- We check for "name or url" in a number of places to alter the source from initial metadata and tarballs
- Once tarball is downloaded, the rest of the PyPI logic takes over

**The original Grayskull command is maintained. Replacing the package name with the GitHub repository URL will generate the recipe.
For example:**

`grayskull pypi https://github.com/confluentinc/confluent-kafka-python
`
@jaimergp

### DIFF
--- a/grayskull/pypi/pypi.py
+++ b/grayskull/pypi/pypi.py
@@ -143,7 +143,7 @@ class PyPi(AbstractRecipeModel):
         return archive_tarball_url
 
     @staticmethod
-    def _download_sdist_pkg(sdist_url: str, dest: str, name: str = None):
+    def _download_sdist_pkg(sdist_url: str, dest: str, name: Optional[str] = None):
         """Download the sdist package
 
         :param sdist_url: sdist url


ForgottenProgramme:
The test it says is failing, is actually passing on my local machine. 
How do I rerun these tests?

marcelotrevisani:
Thank you very much @ForgottenProgramme for you contribution!

I am very impressed by the detailed message you wrote in the PR, pretty nice! Keep it up! :tada: :tada: 

Lets try to get

- https://github.com/conda-incubator/grayskull/pull/245

merged and after that, I will generate a new version with your modifications :)

PR #264, opened 2021-11-16
generic message

ocefpaf:
Now that grayskull supports both pypi and github as sources for python packages we need more "generic messages"

I'm not sure if this is OK but I want to start the conversation. Also, it is odd to issue the command:

```
grayskull pypi gh-repo
```

Maybe we could rename that to:

```
grayskull python any-source
```

or short-circuits to `grayskull gh` and `grayskull pypi` if that makes sense.

### DIFF
--- a/grayskull/__main__.py
+++ b/grayskull/__main__.py
@@ -128,9 +128,7 @@ def main(args=None):
                 is_strict_cf=args.is_strict_conda_forge,
             )
         except requests.exceptions.HTTPError as err:
-            print_msg(
-                f"{Fore.RED}Package seems to be missing.\nException: {err}\n\n"
-            )
+            print_msg(f"{Fore.RED}Package seems to be missing.\nException: {err}\n\n")
             continue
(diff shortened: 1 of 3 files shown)


marcelotrevisani:
> Now that grayskull supports both pypi and github as sources for python packages we need more "generic messages"
> 
> I'm not sure if this is OK but I want to start the conversation. Also, it is odd to issue the command:
> 
> ```
> grayskull pypi gh-repo
> ```
> 
> Maybe we could rename that to:
> 
> ```
> grayskull python any-source
> ```
> 
> or short-circuits to `grayskull gh` and `grayskull pypi` if that makes sense.

Yeah, I think would be nice indeed. But we will need to add a warning that the old way is deprecated, and we might support it for a while but after some time we can drop it
Just black failed, can you fix it please? after that we can merge it :)

marcelotrevisani:
Thanks! :tada: :tada: :tada:

Issue #204, opened 2020-11-22
[FEATURE] Use grayskull with any pypi mirror/server


lucharo:
**Is your feature request related to a problem? Please describe.**
As per my reading and testing of this tool, I do not think it is is possible to make a conda recipe out of **any** pypi mirror just out of the official one (ref to code: https://github.com/conda-incubator/grayskull/blob/d72b59804ab6800d18f1cef40c03ff22565ff2b2/grayskull/pypi/pypi.py#L40). 
**Describe the solution you'd like**
I would like to be able to use `grayskull` to make conda recipes out of any python package listed in any arbitrarily provided pypi mirror or even locally. Perhaps by having a CLI command option:
```bash
grayskull --mirror https://pypi.myenterprise.com <python-package-name>
```

**Describe alternatives you've considered**
Building a conda recipe manually and having undesired behaviours.

marcelotrevisani:
Yes, that is something which is good to have indeed!

For another repositories that would be relatively easy to change, for local/specific paths it will need a bit more of modifications. But it is something doable.

I will add it to my plans, thanks for this request!

lc5415:
Nice, thanks Marcelo! I think this feature will be especially useful in enterprise environments (which is precisely my use case) 😄

lc5415:
Is this line:
https://github.com/conda-incubator/grayskull/blob/d72b59804ab6800d18f1cef40c03ff22565ff2b2/grayskull/pypi/pypi.py#L40

The one actually responsible for `grayskull` looking in pypi.org only? If I changed that one to my entreprise pypi server would it work?

marcelotrevisani:
> Is this line:
> https://github.com/conda-incubator/grayskull/blob/d72b59804ab6800d18f1cef40c03ff22565ff2b2/grayskull/pypi/pypi.py#L40
> 
> The one actually responsible for `grayskull` looking in pypi.org only? If I changed that one to my entreprise pypi server would it work?

It will depend if your enterprise pypi server will return the metadata correctly. If they return it like pypi it will work, otherwise I am afraid that might not work
But it is on my plans to also generate the recipe from sdist packages located anywhere, that will solve your problem as well

lc5415:
Actually, I think it would be really useful to have `grayskull` look in all the channels specified by the `.pypirc` and `.condarc` the user has. What do you think @marcelotrevisani ?

marcelotrevisani:
[...]

PR #274, opened 2022-01-15
Add option --pypi-url option

marcelotrevisani:
Add cli option to change the url that grayskull will remover the pypi metadata


marcelotrevisani:
Partially solves #204

PR #291, opened 2022-02-07
Fix generic_py_ver_to with compatible release

### LINKED ISSUE #290: [BUG] `requires_python ~= 3.6` generates incorrect python requirement
**Describe the bug**
I ran `grayskull pypi wheel-filename` and the python requirement generated was `python >=3.6,<3.7`.
The [setup.cfg](https://github.com/jwodder/wheel-filename/blob/4099de720a66854bcf2b0f02cb4f276fdb2f886a/setup.cfg#L42) includes `python_requires = ~= 3.6`, which is equivalent to `>=3.6,<4`.

**To Reproduce**
1. Run `grayskull pypi wheel-filename`

**Expected behavior**
The python requirement generated should be `>=3.6,<4` (or `>=3.6,==3.*`?)

**Environment:**
 - grayskull --version
1.1.0

**Additional context**
I tracked the issue to the `generic_py_ver_to` function.
Currently `generic_py_ver_to({"requires_python": "~=3.6"}, config) == ">=3.6,<3.7` which is incorrect.

beenje:
This is more a workaround as it's not really correct.
I convert `~=3.6` to `>=3.6`, instead of `>=3.6,<4` (or `>=3.6,==3.*`).

For python versions, we should always have major.minor specified as compatible release and I'm not sure we want to force the upper bound on the major version in conda recipes. At least not for Python 4.
The python 2/3 case was different but... I hope we can start to not care about Python 2...

What do you think?
Is this good enough? Or should we try to do the correct thing?

Fix #290

### DIFF
--- a/grayskull/config.py
+++ b/grayskull/config.py
@@ -78,6 +78,8 @@ class Configuration:
         for op, major, minor in req_python:
             if op == "=":
                 op = "=="
+            elif op == "~=":
+                op = ">="
             if not minor:
                 minor = 0
             for sup_py, is_enabled in py_ver_enabled.items():
--- a/grayskull/strategy/py_base.py
+++ b/grayskull/strategy/py_base.py
@@ -142,7 +142,7 @@ def generic_py_ver_to(
     if not metadata.get("requires_python"):
         return None
     req_python = re.findall(
-        r"([><=!]+)\s*(\d+)(?:\.(\d+))?",
+        r"([~><=!]+)\s*(\d+)(?:\.(\d+))?",
         metadata["requires_python"],
     )
     if not req_python:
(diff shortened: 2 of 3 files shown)


marcelotrevisani:
Did you tested with ``~=3.6.1`` for instance? because that should be ``>=3.6,<3.7``

beenje:
> Did you test with `~=3.6.1` for instance? because that should be `>=3.6,<3.7`

With the current workaround it will be `>=3.6`...
It should really be `>=3.6.1,<3.7`.

That would require quite some change to the implementation as it always returns some `PyVer = namedtuple("PyVer", ["major", "minor"])`.

I've never seen any package requiring a specific patch version of Python (OK, someone could use `~=3.6.0`).
It's a corner case that is probably quite rare.

I'm of course not really satisfied with this argument.
I'm just not sure it's worth the effort. But you decide :-)

beenje:
> Hi @beenje , sorry for the delay, I was quite busy these days, I still need to check one thing, this PR has not been forgotten! :)

No worry! Just let me know if you think it's good enough for now or it should be improved.

marcelotrevisani:
yeah, I think that is fair, thanks for your contribution! :)

PR #293, opened 2022-02-12
Fix local sdist source

beenje:
For local sdist archive, source shall be url and not path.
path is only for directories.

I don't know how I missed that...


marcelotrevisani:
yeah, indeed, thanks for fixing it!

Issue #299, opened 2022-02-16
[BUG] numpy is added with `pin_compatible` and version.


BastianZim:
**Describe the bug**

**To Reproduce**
1. Generate a recipe for `dimod`.
2. Output includes 
  * `- {{ pin_compatible('numpy') }} `
  * `- numpy >=1.19.1,<2.0.0`

**Expected behavior**

Only `- {{ pin_compatible('numpy') }} ` is kept and numpy versions are added to the `host`.

**Outputs**
If applicable, add the output to help explain your problem.

```yaml
{% set name = "dimod" %}
{% set version = "0.10.13" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

build:
  skip: true  # [py<37]
  script: {{ PYTHON }} -m pip install . -vv
  number: 0

source:
  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/dimod-{{ version }}.tar.gz
  sha256: b854a5e633d65b836fc8b90d4fb1ef411843629e3dc49f237e6846060f7b6d38

requirements:
  build:
    - {{ compiler('c') }}
  host:
    - cython
    - numpy
    - pip
    - python
  run:
    - numpy >=1.19.1,<2.0.0
    - pyparsing >=2.4.7,<3.0.0
    - python
    - {{ pin_compatible('numpy') }}

test:
  imports:
    - dimod
  commands:
    - pip check
  requires:
    - pip

about:
  home: https://github.com/dwavesystems/dimod
  summary: A shared API for binary quadratic model samplers.
  license: Apache-2.0
  license_file: LICENSE

extra:
  recipe-maintainers:
    - ADD_YOUR_GITHUB_ID_HERE
```
**Environment:**
https://www.marcelotrevisani.com/grayskull

**Additional context**

Can be tested with `dimod`.

marcelotrevisani:
yeah, that is a problem, gonna fix it, thanks

PR #301, opened 2022-02-17
Fix mixed numpy and pin_compatible

marcelotrevisani:
Fix: #299


marcelotrevisani:
not sure why the test is failing.... it is passing on my local machine, it needs further investigation

BastianZim:
Thank you!

PR #305, opened 2022-02-23
Display warning about license match percentage

### LINKED ISSUE #303: [FEATURE] Add comment to identified license if match is not perfect
`grayskull` will identify licenses by fuzzy match in some cases and then suggest the best result. However, sometimes the fuzzy match scores a bit lower than average because some people _modify_ popular licenses like BSD or MIT and add their own clauses. This kind of change will turn those into a different license altogether, which should be named as `LicenseRef-packagename`.

**Describe the solution you'd like**
I suggest:

* Adding a comment next to the suggested identifier, saying that its match % was lower than usual `license: MIT  # warning! license might have been modified; search match is XX%`.
* If no clear license was found, generate a `LicenseRef-packagename` identifier with a comment next to it like `# license could not be identified, please review`.

**Describe alternatives you've considered**

Erroring out seems excessive.

**Additional context**

A recent [example](https://github.com/gdellaire/BBii-Decon/blob/18563b2bb823977ffa19628c6f7cdb6fe0b35ca3/LICENSE#L20-L22) I found on a conda-forge submission.

ForgottenProgramme:
Fixes #303 (partly)

Grayskull uses rapidfuzz to do the matching between the package license and a set of standard licenses stored in grayskull/license/data. 
I have added a warning to inform the user about how close the 'License Type' (as mentioned in the recipe and displayed in the console) is to the actual license found in the package metadata. 
As mentioned in the linked issue and as discussed with @marcelotrevisani, the next step should be to add a comment about the match percentage within the recipe.


grayskull/license/discovery.py
```diff
@@ -344,6 +345,10 @@ def get_license_type(path_license: str, default: Optional[str] = None) -> Option
     best_match = process.extract(
         license_content, licenses_text, scorer=token_sort_ratio
     )
+    print(
```
marcelotrevisani, line 348: can you replace the print by
```suggestion
    print_msg(
```
please?

marcelotrevisani:
Cool! Thanks a lot for your contribution again! :)
🚀🚀🚀

PR #315, opened 2022-04-06
Update pre-commit config and run black

beenje:
I noticed in PR #314 that pre-commit failed: https://github.com/conda-incubator/grayskull/runs/5850605951?check_suite_focus=true

This is a [known issue](https://github.com/psf/black/issues/2976) with latest version of click and black.
It was fixed in black 22.3.0.

I updated pre-commit config and ran the new version of black.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,15 @@
 exclude: docs
 repos:
 -   repo: https://github.com/psf/black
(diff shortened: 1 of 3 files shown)


marcelotrevisani:
Indeed, thanks for fixing it!

marcelotrevisani (approved):

PR #317, opened 2022-04-22
add test requirements and pytest command

woutdenolf:
This might be too specific so let me know if you have suggestions on how to generalize it.

I have several packages that supports this for testing

```
pip install ewokscore[test]
pytest --pyargs ewokscore
```

This PR adds a CLI option

```
grayskull pypi ewokscore --extras-require-test test
```

It then looks for `"test"` in `extras_require` to add requirements to the `test` section of the recipe.

In addition if `"pytest"` is in the test requirements it adds the command `pytest --pyargs {name}` to the test commands.


grayskull/strategy/py_base.py
```diff
@@ -602,6 +603,13 @@ def get_test_imports(metadata: dict, default: Optional[str] = None) -> List:
     return result
 
 
+def get_test_requirements(metadata: Dict) -> List:
+    try:
+        return metadata["extras_require"][CLIConfig().extras_require_test]
```
marcelotrevisani, line 608: Just to cover a few corner case we might change it to
```suggestion
        return metadata["extras_require"][geattr(CLIConfig(), "extras_require_test", [])]
```

grayskull/strategy/py_base.py
```diff
@@ -602,6 +603,13 @@ def get_test_imports(metadata: dict, default: Optional[str] = None) -> List:
...
+    except KeyError:
```
marcelotrevisani, line 609: or you can add another exception here
```suggestion
    except KeyError, AttributeError:
```

marcelotrevisani:
Cool! Sorry for my delay I have beeing quite buse in the past weeks. 
It looks good to me the general idea, I'm onboard with that. Just a few things that would be nice to add in this PR if you don't mind. Can you add a test for it please? it would be really cool if you add it to cover this new feature.
There is one failing test, I added two suggestions to the PR that will cover those cases, if you add any of them should be fine, the second one adding a new exception might be better as your current implementation you are using the try except block
but I am happy with either implementation :)

marcelotrevisani:
there is just one failing test that need to be fixed, after that it should be good to go

marcelotrevisani (approved):

marcelotrevisani:
Cool! Thanks for your contribution! I will cut a new release after that

PR #344, opened 2022-06-18
robust way of checking whether a package is in a list of requirements

### LINKED ISSUE #343: [BUG] checks of package in list of dependencies
In several places we check whether a package is in the a list of requirements

https://github.com/conda-incubator/grayskull/blob/ade688f1fe548325717b2d2e3ddf1d4d55c1818b/grayskull/strategy/py_base.py#L266

That will fail, for example

```
setup_requires = ['setuptools_scm[toml] >= 3.4.1']
```

woutdenolf:
Closes #343

By introducing the robust check, some tests related to `setuptools-scm` fail. I don't know what the expected result should be. In the code `setuptools-scm` is removed from the requirements but in the tests it is expected to be there.

### DIFF
--- a/grayskull/strategy/py_base.py
+++ b/grayskull/strategy/py_base.py
@@ -25,6 +25,8 @@ from grayskull.utils import (
     merge_dict_of_lists_item,
     merge_list_item,
     origin_is_github,
+    package_in_requirements,
+    remove_package_from_requirements,
     sha256_checksum,
 )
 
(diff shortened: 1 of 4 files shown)


marcelotrevisani:
> Closes #343
> 
> By introducing the robust check, some tests related to `setuptools-scm` fail. I don't know what the expected result should be. In the code `setuptools-scm` is removed from the requirements but in the tests it is expected to be there.

the tests that failed regarding setuptools-scm should pass, because they are a few corner cases where setuptools-scm should be added to the recipe

marcelotrevisani:
> > the tests that failed regarding setuptools-scm should pass, because they are a few corner cases where setuptools-scm should be added to the recipe
> 
> Are you referring to this?
> 
> https://github.com/conda-incubator/grayskull/blob/ade688f1fe548325717b2d2e3ddf1d4d55c1818b/grayskull/strategy/py_base.py#L357
> 
> So in the tests that check whether `setuptools-scm` is in the requirements, `use_scm_version` should the `True`?

if you turn the ``use_scm_version`` to true ``grayskull`` will add the ``setuptools-scm``.
That exists because there are some recipes that people forgot to add ``setuptools-scm`` in the metadata but the presence of that key in the distutils informs that setuptools-scm will be used.

Issue #320, opened 2022-04-26
Create recipe from github tag


marcelotrevisani:
This is currently available: it allows one to use the latest release.

```sh
grayskull pypi https://github.com/confluentinc/confluent-kafka-python
```

:hand:  How can you create a recipe for a specific release version (with a tag preferably)?

:fire: Is there something like the following that could be used?

```sh
grayskull pypi https://github.com/{{ OWNER }}/{{ REPO }} --tag {{ TAG }}
```

Example:

```sh
OWNER="sugatoray"
REPO="genespeak"
TAG="v0.0.9"
grayskull pypi https://github.com/${OWNER}/${REPO} --tag ${TAG}
```

## Notes

I see that the code tries to download the sdist package from this URL.

```sh
https://api.github.com/repos/{{ USER }}/{{ REPO }}/releases/latest
```

According to [GitHub API docs](https://docs.github.com/en/rest/releases/releases#get-a-release-by-tag-name) the following returns the release information with a specific tag.

```sh
https://api.github.com/repos/{{ USER }}/{{ REPO }}/releases/tags/{{ TAG }}
```

### Example

Here we look at the json output returned by GitHub API for genespeak repository. Both of the following two methods will yield the same output (release info for `v0.0.9` as on 2022-04-26).

- https://github.com/sugatoray/genespeak

Release: `latest`

```sh
https://api.github.com/repos/sugatoray/genespeak/releases/latest
```

<details>
<summary><strong>Click to expand and see output</strong></summary>
<p>

```json
{
  "url": "https://api.github.com/repos/sugatoray/genespeak/releases/57632929",
  "assets_url": "https://api.github.com/repos/sugatoray/genespeak/releases/57632929/assets",
  "upload_url": "https://uploads.github.com/repos/sugatoray/genespeak/releases/57632929/assets{?name,label}",
  "html_url": "https://github.com/sugatoray/genespeak/releases/tag/v0.0.9",
  "id": 57632929,
  "author": {
    "login": "sugatoray",
    "id": 10201242,
    "node_id": "MDQ6VXNlcjEwMjAxMjQy",
    "avatar_url": "https://avatars.githubusercontent.com/u/10201242?v=4",
    "gravatar_id": "",
    "url": "https://api.github.com/users/sugatoray",
    "html_url": "https://github.com/sugatoray",
    "followers_url": "https://api.github.com/users/sugatoray/followers",
[...]

PR #346, opened 2022-07-01
From GitHub tag

dillonroach:
Addresses #320 

Hi - I overlooked a contributing.md or the like, so apologies if this isn't in the right place.
I wanted to give the git tags issue a whirl as an intro to this fun project

What this does:
adds a --tag (-t) cli addition for the pypi strategy that then propagates that 'tag' through config and into a new 'verify_github_repo_tag()' in base/github.py when a supplied tag shows up not-None in handle_gh_version()
the verify function looks now at f"/repos{url_parts.path}/git/refs/tags/{tag}" to see if a supplied tag exists - 
if it does exist/match it sets that as version_tag and version in handle_gh_version()

I wasn't certain if this covered the issue, or if the python version identifier also needed to be updated - would appreciate the review; thanks in advance.

### DIFF
--- a/grayskull/__main__.py
+++ b/grayskull/__main__.py
@@ -130,6 +130,13 @@ def main(args=None):
         dest="extras_require_test",
         help="Extra requirements to run tests.",
     )
+    pypi_cmds.add_argument(
+        "--tag",
+        "-t",
+        default=None,
+        dest="github_release_tag",
+        help="If tag is specified, grayskull will build from given release tag, rather than latest.",
+    )
 
     args = parser.parse_args(args)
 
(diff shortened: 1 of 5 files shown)


dillonroach:
Apologies - managed to miss the flake8 error in the precommit pass last time - not sure how it slipped by the last pass.

Provided the PR looks good, I don't think it's WIP anymore.

marcelotrevisani (approved):

marcelotrevisani:
cool, thanks for your contribution! :)

sugatoray:
👏👏 Wonderful! 

👉 Is the `README.md` also updated with an example of how to use the new feature?

cc: @marcelotrevisani @dillonroach

PR #365, opened 2022-08-06
Use "compatible release" operator ~=

### LINKED ISSUE #348: [BUG] "Using .* with relational operator is superfluous and deprecated"
Consider the package [`dbt-core`](https://pypi.org/project/dbt-core/#files). The `install_requires` contains

```python
        "dbt-extractor~=0.4.1",
```

[Running Grayskull](https://marcelotrevisani.info:8000/recipe/?name=dbt-core&version=1.1.1) this gets translated into

```yaml
    - dbt-extractor >=0.4.1,==0.4.*
```

Then running boa with `conda mambabuild -c conda-forge .` or `conda build -c conda-forge .` produces the warning

```
WARNING:conda.models.version:Using .* with relational operator is superfluous and deprecated and will be removed in a future version of conda. Your spec was 0.4.*, but conda is ignoring the .* and treating it as 0.4
```

This results in the unsatisfiable dependency `>=0.4.1,==0.4` since `==0.4` seems to imply 0.4.0 and excludes 0.4.1.

Interestingly, if I change it to a single equals (`>=0.4.1,=0.4.*`) then it doesn't complain and seems to work as intended.

maresb:
Building such recipes requires conda-build >=3.20.5

Closes #348

### DIFF
--- a/tests/test_pypi.py
+++ b/tests/test_pypi.py
@@ -423,7 +423,7 @@ def test_normalize_requirements_list():
     config = Configuration(name="pytest")
     requirements = ["pytest ~=5.3.2", "pyqt5"]
     requirements = set(normalize_requirements_list(requirements, config))
-    expected = {"pytest >=5.3.2,==5.3.*", "pyqt"}
+    expected = {"pytest ~=5.3.2", "pyqt"}
     assert requirements == expected
 
 
@@ -1170,7 +1170,7 @@ def test_get_test_imports_clean_modules():
 
 
 def test_ensure_pep440():
-    assert ensure_pep440("pytest ~=5.3.2") == "pytest >=5.3.2,==5.3.*"
+    assert ensure_pep440("pytest ~=5.3.2") == "pytest ~=5.3.2"
 
 
 def test_pep440_recipe():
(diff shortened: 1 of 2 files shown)


maresb:
CC @BastianZim

tests/test_pypi.py
```diff
@@ -423,7 +423,7 @@ def test_normalize_requirements_list():
     config = Configuration(name="pytest")
     requirements = ["pytest ~=5.3.2", "pyqt5"]
     requirements = set(normalize_requirements_list(requirements, config))
-    expected = {"pytest >=5.3.2,==5.3.*", "pyqt"}
+    expected = {"pytest ~=5.3.2", "pyqt"}
```
marcelotrevisani, line 426: cc @ocefpaf 
what do you think? should we revert to use ``~=`` as conda-build supports it https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#build-version-spec

maresb: Actually the `~=` operator is not documented in that link, but it is supported. See https://github.com/conda/conda-build/issues/4553.

ocefpaf: I don't have a strong opinion. My take is, if conda-build supports it and it is a python standard we should too. In the past some folks complained that parts of the bot (I don't really remember) choked on this syntax. Maybe it is worth asking in the core gitter is someone remembers what was wrong with recipes that used that syntax.

maresb: Thanks @ocefpaf for the feedback. Regarding potential problems, it is necessary to use `conda-build >=3.20.5`. We have already added this to Grayskull's `run_constrained`.

It would be great if you could ask in the core gitter to check if there are any further known issues with `~=`. Thanks so much for looking into this!

marcelotrevisani: cool, lets get it merged then! :tada:

marcelotrevisani:
lets get this moving, I think that is fine at the current state and ``~=`` will avoid problems with versions when trying  to convert it

maresb:
Sounds good. If we learn something bad about `~=` we can always revert it before the next release.

PR #368, opened 2022-08-10
Fix "compatible release" operator

maresb:
According to https://github.com/conda-incubator/grayskull/issues/348#issuecomment-1210348234, the currently-merged solution will break back-compatibility. Let's implement another solution instead.

(Hopefully I can take care of it this evening.)

### DIFF
--- a/grayskull/strategy/py_base.py
+++ b/grayskull/strategy/py_base.py
@@ -753,4 +753,4 @@ def next_incompatible_version(version: str) -> str:
         )
     release = previous_release[:-2] + (previous_release[-2] + 1,)
     release_str = ".".join(str(r) for r in release)
-    return canonicalize_version(f"{epoch}!{release_str}a")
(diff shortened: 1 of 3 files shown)


maresb:
I want to write code which is as reasonably correct and robust as practical. According to the [Conda docs](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/pkg-specs.html#predictable-version-ordering), `dev`<`a`:
image

And also the `packaging` package agrees:
```python
>>> from packaging.version import Version
>>> a = Version("1.0.0a0")
>>> dev = Version("1.0.0dev0")
>>> dev<a
True
>>> a<dev
False
```

Thus I'm replacing `a` with `dev`. Any objections?

marcelotrevisani:
that is perfect to me

marcelotrevisani:
thanks for your contribution again! :tada: :tada: :tada:

maresb:
Thanks so much for seeing this through with me!!! :smile:

PR #371, opened 2022-08-13
Make rm_duplicated_deps robust to mixed dashes and underscores

maresb:
This fixes a fairly obscure edge case, and I don't have any real instances in mind, but I added a test case which would otherwise fail.

(Previously, with input `["my_craZy-pkg", "my-crazy-pkg"]`, the second package is not recognized as duplicate since it only checks the previous entries for `my-crazy-pkg` and `my_crazy_pkg`.)


marcelotrevisani:
gotcha, indeed is a very specific corner case, thanks for adding the tests!

marcelotrevisani (approved):

PR #375, opened 2022-08-21
Maintain dependency order

### LINKED ISSUE #259: Maintaining dependency order

When looking at diffs of dependencies, sometimes they are hard to reason about due to reordering of the dependencies. Additionally some dependencies (like `python` tend to get listed first to make it easier to see they are there (and not forgotten) with any relevant version constraints.

**Describe the solution you'd like**

It would help if grayskull was able to maintain the dependency order as much as possible. This would keep diffs more compact and making it easier to catch any errors (like C/C++ dependencies that got dropped).

**Describe alternatives you've considered**

One alternative instead of following the recipe order would be to follow the `setup.py` order. As maintainers are often comparing the two, this actually makes that comparison easier. Though if the recipe is doing something different, this might be off for the first PR still

Feel free to suggest alternatives.

**Additional context**

Have made and looked at a few PRs with Grayskull recently and this behavior has resulted in some confusion.

maresb:
Currently dependencies are alphabetically sorted (with the exception of Python which goes at the top). This can be confusing when comparing Grayskull's output to the upstream repository. This was previously requested in #259.

In general I find alphabetization good since it lowers entropy. But as far as Grayskull is concerned, it feels to me like alphabetization would be better done upstream. But I could understand people wanting to preserve the alphabetization behavior.

In case we decide against alphabetization, I hope that we could still merge all but the last commit which flips the switch.

Closes #259

### DIFF
--- a/grayskull/strategy/pypi.py
+++ b/grayskull/strategy/pypi.py
@@ -543,8 +543,8 @@ def extract_requirements(metadata: dict, config, recipe) -> Dict[str, List[str]]
     return result
 
 
-def sort_reqs(reqs: Iterable[str]) -> List[str]:
-    """Sort requirements. Put python first, then sort alphabetically."""
+def sort_reqs(reqs: Iterable[str], alphabetize: bool = True) -> List[str]:
+    """Sort requirements. Put python first, then optionally sort alphabetically."""
     reqs_list = list(reqs)
 
     def is_python(req: str) -> bool:
@@ -552,7 +552,9 @@ def sort_reqs(reqs: Iterable[str]) -> List[str]:
 
     python_reqs = [req for req in reqs_list if is_python(req)]
     non_python_reqs = [req for req in reqs_list if not is_python(req)]
-    result = python_reqs + sorted(non_python_reqs)
+    if alphabetize:
+        non_python_reqs.sort()
+    result = python_reqs + non_python_reqs
     return result
 
 
--- a/grayskull/strategy/pypi.py
+++ b/grayskull/strategy/pypi.py
@@ -543,7 +543,7 @@ def extract_requirements(metadata: dict, config, recipe) -> Dict[str, List[str]]
     return result
 
 
-def sort_reqs(reqs: Iterable[str], alphabetize: bool = True) -> List[str]:
+def sort_reqs(reqs: Iterable[str], alphabetize: bool = False) -> List[str]:
     """Sort requirements. Put python first, then optionally sort alphabetically."""
     reqs_list = list(reqs)
 
--- a/grayskull/utils.py
+++ b/grayskull/utils.py
@@ -118,20 +118,27 @@ def string_similarity(a, b):
 def rm_duplicated_deps(all_requirements: Union[list, set, None]) -> Optional[list]:
     if not all_requirements:
         return None
-    new_value = []
-
     # Keep track of requirements which have already been added to the list.
-    # Here we store a canonicalized version of the requirement: lowercase,
-    # and underscores converted to dashes.
-    added_reqs = set()
+    # The key is a canonicalized version of the requirement: lowercase,
+    # and underscores converted to dashes. The value is the requirement itself,
+    # as it should be added.
+    # (This is order-preserving since dicts are ordered by first insertion.)
+    new_reqs: dict[str, str] = {}
 
     for dep in all_requirements:
         canonicalized = dep.replace("_", "-").lower()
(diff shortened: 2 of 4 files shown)


marcelotrevisani (approved):
Cool! Thanks for the contribution again! :)

maresb:
Yay!!! Thank you!

xylar:
I like this idea but with `pyproject.toml` files it seems like the dependencies are resulting in exactly the opposite order of what they were in the original.  I can post an issue about this.

marcelotrevisani:
> I like this idea but with `pyproject.toml` files it seems like the dependencies are resulting in exactly the opposite order of what they were in the original. I can post an issue about this.

yes, please do! But grayskull does not support pyproject.toml yet, I have plans to add this support in the upcoming months. But it is good to be aware of any case scenario

maresb:
@xylar, for my reference could you send me an example package where you're seeing the reverse ordering?

maresb:
@xylar, perhaps this is fixed by https://github.com/conda-incubator/grayskull/pull/379?

@marcelotrevisani, would it be possible to release v1.8.1 with this fix? Currently v1.8.0 is randomizing the dependency order. :man_facepalming:

marcelotrevisani:
Sure, let me do it

maresb:
Wow, awesome!!! You're fast! :laughing:

xylar:
@maresb, maybe we should move the conversation to #381.  I installed the latest from `main` and I still see the backwards order.

> yes, please do! But grayskull does not support pyproject.toml yet, I have plans to add this support in the upcoming months. But it is good to be aware of any case scenario

@marcelotrevisani, okay, but grayskull works fine in the sense of not giving any error message when I run it for poetry:
```
$ grayskull pypi poetry=1.2.0







#### Initializing recipe for poetry=1.2.0 (pypi) ####

Recovering metadata from pypi...
Starting the download of the sdist package poetry
poetry 100% Time:  0:00:01 527.8 KiB/s|#######################################|
Recovering information from setup.py
Executing injected distutils...
Recovering metadata from setup.cfg
No data was recovered from setup.py. Forcing to execute the setup.py as script
Recovering metadata from setup.cfg
Checking >> poetry-core 100% |#########################|[Elapsed Time: 0:00:09]
Recovering license info from spdx.org ...
Matching license file with database from Grayskull...
Match percentage of the license is 97%. Low match percentage could mean that the license was modified.
Matching license file with database from Grayskull...
Match percentage of the license is 97%. Low match percentage could mean that the license was modified.
License type: MIT
License file: ['LICENSE']
License type: MIT
License file: ['LICENSE', 'tests/fixtures/with-include/LICENSE']
Build requirements:
  <none>
Host requirements:
  - python
  - ansible
  - pip
Run requirements:
  - python
  - dulwich >=0.20.44,<0.21.0
  - urllib3 >=1.26.0,<2.0.0
  - xattr >=0.9.7,<0.10.0  # [darwin]
  - virtualenv
  - tomlkit >=0.11.1,<1.0.0,!=0.11.2,!=0.11.3
  - shellingham >=1.5,<2.0
  - requests-toolbelt >=0.9.1,<0.10.0
  - requests >=2.18,<3.0
  - platformdirs >=2.5.2,<3.0.0
  - pkginfo >=1.5,<2.0
  - pexpect >=4.7.0,<5.0.0
  - packaging >=20.4
  - keyring >=21.2.0
  - jsonschema >=4.10.0,<5.0.0
  - importlib-metadata >=4.4,<5.0  # [py<310]
  - html5lib >=1.0,<2.0
  - crashtest >=0.3.0,<0.4.0
  - cleo >=1.0.0a5,<2.0.0
  - cachy >=0.3.0,<0.4.0
  - cachecontrol >=0.12.9,<0.13.0
  - poetry-plugin-export >=1.0.6,<2.0.0
  - poetry-core ==1.1.0

RED: Missing packages
GREEN: Packages available on conda-forge
Using default recipe maintainer: AddYourGitHubIdHere

Maintainers:
   - AddYourGitHubIdHere

#### Recipe generated on /home/xylar/Desktop for poetry=1.2.0 ####
```
There's no `setup.cfg` or `setup.py` that I can see in the source:
https://files.pythonhosted.org/packages/d6/75/2603419dcfafb3e777b304c9cb2813a8a4090db04356b750a662663c3c01/poetry-1.2.0.tar.gz
so I have to assume that info is coming from `pyproject.toml` or perhaps one of the other files that is there?

PR #378, opened 2022-08-24
Add exceptiongroup to avoid_selector

BastianZim:
It is mentioned here that it is save to be installed under all versions: https://github.com/agronholm/exceptiongroup

### DIFF
--- a/grayskull/strategy/config.yaml
+++ b/grayskull/strategy/config.yaml
@@ -450,3 +450,8 @@ pywin32:
   conda_forge: pywin32
   import_name: pywin
   avoid_selector: true
+
+exceptiongroup:
+  conda_forge: exceptiongroup
+  import_name: exceptiongroup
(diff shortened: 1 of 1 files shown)


grayskull/strategy/config.yaml
```diff
@@ -450,3 +450,8 @@ pywin32:
   conda_forge: pywin32
```
BastianZim, line 450: Shouldn't this be pywin32-on-windows by the way: https://github.com/conda-forge/pywin32-on-windows-feedstock

marcelotrevisani: indeed! can you fix that in this PR as well please?

BastianZim: Done! :)

marcelotrevisani (approved):
thanks for that! :)

PR #379, opened 2022-08-28
Fix scrambled dependency order

maresb:
The order of dependencies is being randomized because a list is being converted to a set and back. This fixes that.


marcelotrevisani (approved):

marcelotrevisani:
Cool! Thanks again for that! :)

PR #386, opened 2022-08-31
use a string for a single `license_file`

woutdenolf:


marcelotrevisani (approved):
at the end of the day that is more a matter of taste than anything, but if they are complaining we can adjust it, no worries. Thanks for that! :)

PR #389, opened 2022-09-01
Fix issue 388

### LINKED ISSUE #388: [BUG] names of multiple outputs vs package name
https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml

```
package:
    name: matplotlib-suite

outputs:
    name: matplotlib-base  # has files

    name: matplotlib  # has no files
```

However when I try this myself for another package, none of the outputs have files

https://github.com/conda-forge/pypushflow-feedstock/blob/main/recipe/meta.yaml

```
package:
    name: pypushflow-meta

outputs:
    name: pypushflow  # has no files

    name: pypushflow-mx  # has no files
```

I traced through the `conda build` code and found this:

https://github.com/conda/conda-build/blob/432a9e1b41a3dec8f95a7556632f9a93fdf029fd/conda_build/build.py#L2274

So it seems that to have files bundled with an output, the output name must be equal to the package name.

In that case I don't understand how `matplotlib`.

In any case, `grayskull` should probably produce this

```
package:
    name: mypackage

outputs:
    name: mypackage  # has files

    name: mypackage-extra  # has no files
```

woutdenolf:
Closes #388

Handle pip extra's as follows:

```
package:
    name: mypackage
    ....

requirements:
    ....

build:
    ....

test:
    ....

outputs:
    - name: mypackage
    - name: mypackage-option1
        - requirements:
              pypushflow =={{ version }}
              ....
        - test:
              ....
        - build:
            - noarch
    - name: mypackage-option2
        - requirements:
              pypushflow =={{ version }}
              ....
        - test:
              ....
        - build:
            - noarch
```

The `mypackage` inherits all top-level sections (requirements, build, test) and will contain the actual distribution.

The `mypackage-option1` and `mypackage-option2` outputs do NOT inherit top-level sections (although package version and build number seem to be inherited). The build does not contain a script so these conda packages will only add extra requirements (which is the goal).

### DIFF
--- a/grayskull/strategy/pypi.py
+++ b/grayskull/strategy/pypi.py
@@ -421,7 +421,9 @@ def get_metadata(recipe, config) -> dict:
             for option, req_list in optional_requirements.items():
                 req_section = dict(output_req_section)
                 req_section["run"] = list(req_section["run"])
-                req_section["run"].append(f"{name} =={{{{ version }}}}")
+                req_section["run"].append(
+                    f"{{{{ pin_subpackage('{name}', exact=True) }}}}"
+                )
                 req_section["run"].extend(req_list)
                 output = {
                     "name": f"{name}-{option}",
--- a/grayskull/strategy/pypi.py
+++ b/grayskull/strategy/pypi.py
@@ -408,30 +408,28 @@ def get_metadata(recipe, config) -> dict:
     outputs = []
     if optional_requirements:
         if config.extras_require_split:
+            # First output is the package itself. Inherits all top-level
+            # recipe sections like 'requirements', 'build', 'test'.
+            outputs.append({"name": name})
+
+            # Other outputs are empty and only add extra dependencies. Does
+            # not inherit any top-level section.
             output_req_section = dict(requirements_section)
-            requirements_section = {
-                k: v for k, v in requirements_section.items() if k == "host"
-            }
-            outputs.append(
-                {
-                    "name": name,
-                    "requirements": dict(output_req_section),
-                }
-            )
             output_req_section["run"] = [
-                s for s in output_req_section["run"] if "python" in s
+                s for s in output_req_section.get("run", []) if "python" in s
             ]
             for option, req_list in optional_requirements.items():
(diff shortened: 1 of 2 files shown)


marcelotrevisani:
I think for the package inside of run requirements it should be ``pin_subpackage('package_name', exact=True)``

it cannot be just the version because imagine that we have different builds of the package for the same version, it can get any of the other builds if we specify just the version

marcelotrevisani:
so for this case it should be something like
```
package:
    name: mypackage
    ....

requirements:
    ....

build:
    ....

test:
    ....

outputs:
    - name: mypackage
    - name: mypackage-option1
        requirements:
          run:
              - {{ pin_subpackage('mypackage', exact=True) }}
              ....
        - test:
              ....
        - build:
            - noarch
    - name: mypackage-option2
        requirements:
          run:
              - {{ pin_subpackage('mypackage', exact=True) }}
              ....
        - test:
              ....
        - build:
            - noarch
```

marcelotrevisani:
@ocefpaf could you take a look as well please?

grayskull/strategy/pypi.py
```diff
@@ -408,30 +408,30 @@ def get_metadata(recipe, config) -> dict:
...
-                req_section["run"].append(f"{name} =={{{{ version }}}}")
-                req_section["run"].extend(req_list)
-                outputs.append(
-                    {
-                        "name": f"{name}-{option}",
-                        "requirements": req_section,
-                    }
+                req_section["run"] = list(req_section["run"])
+                req_section["run"].append(
+                    f"{{{{ pin_subpackage('{name}', exact=True) }}}}"
```
woutdenolf, line 425: @marcelotrevisani Indeed better. Thanks!

marcelotrevisani (approved):
Cool! Thanks for that again! that is a nice addition!

ocefpaf:
> @ocefpaf could you take a look as well please?

Sorry I was away last week. This looks awesome BTW! Great addition to grayskull!

PR #392, opened 2022-09-05
squeeze out None licenses

ocefpaf:
Grayskull crashes when one of the licenses is None in the `all_license_name` list. This is probably not the best solution but I thought I should try a PR instead of just opening issues ;-p


marcelotrevisani:
Cool! Thanks for that! :)

marcelotrevisani (approved):

PR #393, opened 2022-09-05
add scikit-learn to name mapping

ocefpaf:
Folks can declare either scikit-learn or sklearn on requriements.txt. I did not fix the final name in the recipe yet. Trying to figure out where that goes.

Using `grayskull pypi --strict-conda-forge pymyami` to test this out.

### DIFF
--- a/grayskull/strategy/config.yaml
+++ b/grayskull/strategy/config.yaml
@@ -455,3 +455,7 @@ exceptiongroup:
   conda_forge: exceptiongroup
   import_name: exceptiongroup
   avoid_selector: true
+
+scikit-learn:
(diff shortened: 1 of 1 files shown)


marcelotrevisani:
that is still as "draft", should we integrate it? @ocefpaf

ocefpaf:
> that is still as "draft", should we integrate it? @ocefpaf

Sorry. I completely forgot about this one. I believe we can merge this and I'll follow up on another PR if this is not enough to find sklearn in the requirements as scikit-learn.

marcelotrevisani:
cool, let me merge it then :)

Issue #322, opened 2022-05-05
[FEATURE] add the items from `build-system/requires` of pyproject.toml into `requirements/host`


fsmaibrgm:
**Is your feature request related to a problem? Please describe.**
The pyproject.toml file has a section `build-system/requires` that provides the same kind of information than the option `setup_requires` into a setup.py file.
Those dependencies should be pasted into the `requirements/host` section of the meta.yaml file.

It works as expected for `setup_requires` into a setup.py but the items are ignored for `build-system/requires` into a pyproject.toml.

**Describe the solution you'd like**

The items listed in `build-system/requires` into a pyproject.toml should appear in  the `requirements/host` section of the meta.yaml file.

**Describe alternatives you've considered**
Doing it by hand.

marcelotrevisani:
that is related to https://github.com/conda-incubator/grayskull/issues/45

dillonroach:
I'm going to take some time and give this and 45 a whirl - just wanted to put my name in the hat so I don't duplicate effort from somebody else/vice-versa

marcelotrevisani:
> I'm going to take some time and give this and 45 a whirl - just wanted to put my name in the hat so I don't duplicate effort from somebody else/vice-versa

Please, feel free to tackle that, :)

PR #394, opened 2022-09-19
[WIP] add TOML support

dillonroach:
Addresses #322
(using a notebook as scratch for a moment, using multiple toml files as stand-ins for previously parsed requirements from other source - will go into proper code place with a few questions resolved)

- [x] general toml import into dict
- [x] simple parse of requirements
- [x] split of packages/version str such that get_name_version_from_requires_dist plays nice
- [ ] resolve package version mis-matches (most/least permissive? take one source as truth? allow user input at cli to toggle?)
- [ ] integrate into package
- [ ] tests

### DIFF
--- a/Untitled.ipynb
+++ b/Untitled.ipynb
@@ -1,0 +1,252 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "f7b8e3a0-9414-44e9-8ef5-d9a71132a8ac",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import tomli\n",
+    "import re\n",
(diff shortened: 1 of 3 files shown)


marcelotrevisani:
> @marcelotrevisani - for the case where toml version pinnings don't match with previously built dicts, what's your preferred way to resolve? take previous/new info as truth and overwrite? overwrite with warning? cli input to swap between different behaviors? thanks

I believe we can give priority for the toml files, as in theory people are moving towards toml files and it should be better maintained. However, if you can also raise an warning when this happens, that would be perfect!

marcelotrevisani:
Thank you for your effort @dillonroach , as we already added the initial support for toml in https://github.com/conda-incubator/grayskull/pull/425
I believe we can close it for now, but feel encourage to contribute to grayskull :)
There are more things to be improved in the toml support

PR #413, opened 2022-10-31
Add some underscore-delimited packages

duncanmmacleod:
This PR adds some `config.yaml` entries for packages that have underscores in their names.


marcelotrevisani (approved):
Thanks for adding this!

PR #445, opened 2023-02-09
Move project metadata to pyproject.toml

mwtoews:
This PR moves mostly static project metadata to `pyproject.toml` as described by [PEP 621](https://peps.python.org/pep-0621/) and supported by [setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

This is a close 1:1 conversion of the project metadata, with a few minor alterations:

- Add a keyword
- Remove trailing full stop from description
- Alpha-sort dependency lists
- Change find parameters from "exclude" to "include"
- The default for `include-package-data` is true for pyproject.toml configurations, so this doesn't need to be specified (but it's harmless to add, if wanted)
- Package data is not necessary, these files are included in the wheel

Other changes:

- Move `[build-system]` to top, where it's usually found, remove "wheel"
- Use `setuptools_scm[toml]` and fix typo in `[tool.setuptools_scm]` to enable the configuration from pyproject.toml (I noticed this was repeated in setup.py)
- Use `build` to build sdist and wheel packages rather than calling setup.py directly (see [Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html))
- Run twine check before upload
- Keep setup.py with repeated "name" for now, as this is expected by some tools (e.g. GitHub dependency graph)


marcelotrevisani:
Thanks for this addition @mwtoews ! Much appreciated!

marcelotrevisani (approved):

PR #446, opened 2023-02-15
Misc: correct license, 3.8+ badge, upgrade publish python-version

mwtoews:
This PR has three minor changes:

- Correction to the license from MIT to Apache-2.0
- Update the Python version badge on README.md to 3.8+
- Upgrade the Python version in the publish action to '3.x' which should use the latest stable version. (While this package can only be installed on Python 3.8, it can be built by other versions, including older releases). Also upgrade the checkout and setup-python actions to their "latest" ones.


pyproject.toml
```diff
@@ -13,7 +13,7 @@ authors = [
 description = "Project to generate recipes for conda packages"
 readme = "README.md"
 keywords = ["conda"]
-license = {text = "MIT"}
+license = {text = "Apache-2.0"}
```
marcelotrevisani, line 16: Thank you! I forgot to update this part! grayskull used to be MIT, but I changed it to Apache-2.0

marcelotrevisani (approved):

PR #450, opened 2023-03-05
Delete dead check_pkgs_availability

maresb:
I was looking through the source, and as far as I can tell, `check_pkgs_availability` is not used anywhere. Perhaps it can be removed?


marcelotrevisani (approved):

marcelotrevisani:
Thanks for that, indeed, that function is not necessary anymore

PR #472, opened 2023-05-19
add name extraction from pyproject.toml for github repos

marcelotrevisani:

### DIFF
--- a/grayskull/strategy/py_base.py
+++ b/grayskull/strategy/py_base.py
@@ -707,6 +707,7 @@ def merge_setup_toml_metadata(setup_metadata: dict, pyproject_metadata: dict) ->
     setup_metadata = defaultdict(dict, setup_metadata)
     if not pyproject_metadata:
         return setup_metadata
+    setup_metadata["name"] = pyproject_metadata["name"]
     if pyproject_metadata["about"]["license"]:
         setup_metadata["license"] = pyproject_metadata["about"]["license"]
     if pyproject_metadata["about"]["summary"]:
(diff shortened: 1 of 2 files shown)


grayskull/strategy/py_toml.py
```diff
@@ -259,6 +259,7 @@ def get_all_toml_info(path_toml: Union[Path, str]) -> dict:
         metadata["about"]["dev_url"] = all_urls.get("Source", None)
         metadata["about"]["home"] = all_urls.get("Homepage", None)
     metadata["about"]["summary"] = toml_project.get("description")
+    metadata["name"] = toml_project.get("name")
```
marcelotrevisani, line 262:
```suggestion
    metadata["name"] = metadata.get("name") or toml_project.get("name")
```

grayskull/strategy/py_base.py
```diff
@@ -707,6 +707,7 @@ def merge_setup_toml_metadata(setup_metadata: dict, pyproject_metadata: dict) ->
     setup_metadata = defaultdict(dict, setup_metadata)
     if not pyproject_metadata:
         return setup_metadata
+    setup_metadata["name"] = pyproject_metadata["name"]
```
marcelotrevisani, line 710:
```suggestion
    setup_metadata["name"] = setup_metadata.get("name") or pyproject_metadata["name"]
```

marcelotrevisani:
It would be nice to just create a test for it to cover this case.
Do you want to work on this @scottstanie ? or do you prefer that I take it over?

scottstanie:
If you have time to get to making a test, then I certainly don't mind and it would be great! Otherwise it might take me a little while to get to one.

PR #509, opened 2023-11-17
Add (`python-`)`build` to quirks

xylar:
The conda-forge package [python-build](https://github.com/conda-forge/python-build-feedstock) is the PyPI package [build](https://pypi.org/project/build/) (and not the PyPI package [python-build](https://pypi.org/project/python-build/), which hasn't been updated in 7 years). There is an archived feedstock on conda-forge for [build](https://github.com/conda-forge/build-feedstock) that we do not want to be using.


xylar:
This came up in:
https://github.com/conda-forge/poetry-feedstock/pull/95

PR #518, opened 2024-01-04
prototype implementation of interpreting PEP725 metadata

msarahan:
In https://discuss.python.org/t/pep-725-specifying-external-dependencies-in-pyproject-toml/31888/40, I offered to prototype an implementation of PEP725 for Grayskull. Here we are!

PEP 725 is the PEP to allow expression of external dependencies in pyproject.toml. It basically boils down to a grand scheme to unify a translation table across the many ecosystems. This PR should be interpreted in the context of the discussion at https://discuss.python.org/t/pep-725-specifying-external-dependencies-in-pyproject-toml/31888

This prototype demonstrates the name mapping, but this won't be a real implementation until a definitive name map is created for conda-forge (and perhaps elsewhere).

I have taken the liberty of consolidating test_flit.py and test_poetry.py into test_py_toml.py, because I found the difference between the implementation in the one file (py_toml.py) vs the split-up test implementation to make things harder to find.

CC @rgommers

Update: after all revisions and fumbling, the design here is:

1. [extract external dependencies from pyproject.toml](https://github.com/conda/grayskull/pull/518/files#diff-fb6b818bcb86e6ebb8733757d7942b0ca48568ee9b0f356de25c5e89fd4e1617R267)
2. [Merge these external dependencies with setup.py/setup.cfg metadata. The build dependencies don't have an equivalent in setup.py/setup.cfg, so we make a separate placeholder for them.](https://github.com/conda/grayskull/pull/518/files#diff-4979c2f9f40da9b20e55dbfe2dece4e1ab40a416a6ded5a8f63b11d9195e110cR744)
3. [get the merged dependencies for the conda recipe requirements](https://github.com/conda/grayskull/pull/518/files#diff-920d6d6f47cb26975a6c1349bae92c12c6db11ed9f46c63a29162c19926e7ccbR560)
4. If any PURL patterns remain in the generated recipe, they are [highlighted in yellow](https://github.com/conda/grayskull/pull/518/files#diff-890f06af677cf1c1c1892f303df4f04041a602ff497012a80d63f0c914d5739aR82). This yellow color is [added to the key]( https://github.com/conda/grayskull/pull/518/files#diff-890f06af677cf1c1c1892f303df4f04041a602ff497012a80d63f0c914d5739aR116)

I think we might want to set a non-zero return code if there are any dependencies not found or PURLs not mapped, but that would be a behavior change. I did not want to get into that with this PR.

### DIFF
--- a/grayskull/strategy/py_base.py
+++ b/grayskull/strategy/py_base.py
@@ -27,6 +27,7 @@ from grayskull.config import Configuration
 from grayskull.license.discovery import ShortLicense, search_license_file
 from grayskull.strategy.py_toml import get_all_toml_info
 from grayskull.utils import (
+    RE_PEP725_PURL,
     PyVer,
     get_vendored_dependencies,
     merge_dict_of_lists_item,
@@ -546,10 +547,12 @@ def update_requirements_with_pin(requirements: dict):
         return [p for p in list_pkgs if pkg != p.strip().split(" ", 1)[0]]
 
     for pkg in requirements["host"]:
-        pkg_name = RE_DEPS_NAME.match(pkg).group(0)
-        if pkg_name in PIN_PKG_COMPILER.keys():
-            requirements["run"] = clean_list_pkg(pkg_name, requirements["run"])
-            requirements["run"].append(PIN_PKG_COMPILER[pkg_name])
+        pkg_name_match = RE_DEPS_NAME.match(pkg)
+        if pkg_name_match:
+            pkg_name = pkg_name_match.group(0)
+            if pkg_name in PIN_PKG_COMPILER.keys():
+                requirements["run"] = clean_list_pkg(pkg_name, requirements["run"])
+                requirements["run"].append(PIN_PKG_COMPILER[pkg_name])
 
 
 def discover_license(metadata: dict) -> List[ShortLicense]:
@@ -733,6 +736,14 @@ def merge_setup_toml_metadata(setup_metadata: dict, pyproject_metadata: dict) ->
             setup_metadata.get("install_requires", []),
             pyproject_metadata["requirements"]["run"],
         )
+    # this is not a valid setup_metadata field, but we abuse it to pass it
+    # through to the conda recipe generator downstream. It's because setup.py
+    # does not have a notion of build vs. host requirements. It only has
+    # equivalents to host and run.
+    if pyproject_metadata["requirements"]["build"]:
+        setup_metadata["__build_requirements_placeholder"] = pyproject_metadata[
+            "requirements"
+        ]["build"]
     if pyproject_metadata["requirements"]["run_constrained"]:
         setup_metadata["requirements_run_constrained"] = pyproject_metadata[
             "requirements"
@@ -802,9 +813,8 @@ def ensure_pep440_in_req_list(list_req: List[str]) -> List[str]:
 
 
 def split_deps(deps: str) -> List[str]:
-    deps = deps.split(",")
     result = []
-    for d in deps:
+    for d in deps.split(","):
         constrain = ""
         for val in re.split(r"([><!=~^]+)", d):
             if not val:
@@ -817,9 +827,10 @@ def split_deps(deps: str) -> List[str]:
 
 
 def ensure_pep440(pkg: str) -> str:
-    if not pkg:
+    if not pkg or RE_PEP725_PURL.match(pkg):
         return pkg
-    if pkg.strip().startswith("<{") or pkg.strip().startswith("{{"):
+    pkg = pkg.strip()
+    if any([pkg.startswith(pattern) for pattern in ("<{", "{{")]):
         return pkg
     split_pkg = pkg.strip().split(" ")
     if len(split_pkg) <= 1:
--- a/grayskull/cli/stdout.py
+++ b/grayskull/cli/stdout.py
@@ -8,6 +8,7 @@ from colorama import Fore, Style
 
 from grayskull.base.pkg_info import is_pkg_available
 from grayskull.cli import WIDGET_BAR_DOWNLOAD, CLIConfig
+from grayskull.utils import RE_PEP725_PURL
 
 
 def print_msg(msg: str):
@@ -78,6 +79,11 @@ def print_requirements(
                 pkg_name = pkg.replace("<{", "{{")
                 options = ""
                 colour = Fore.GREEN
+            elif RE_PEP725_PURL.match(pkg):
+                pkg_name = pkg
+                options = ""
+                all_missing_deps.add(pkg)
+                colour = Fore.YELLOW
             elif search_result:
                 pkg_name, options = search_result.groups()
                 if is_pkg_available(pkg_name):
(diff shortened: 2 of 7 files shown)


rgommers:
Thanks @msarahan! This looks nicely straightforward.

I tried some initial testing, but am running into an unrelated issue with grayskull not understanding non-distutils packages. E.g., for numpy and scipy I see:
```
Checking for pyproject.toml
pyproject.toml found in /tmp/grayskull-numpy-e0nj9egc/numpy-2.0.0.dev0/pyproject.toml
Recovering information from setup.py
Executing injected distutils...
...
```
Followed by an exception (for scipy) or it trying to build and run Cython somehow and crashing (for numpy).

I don't have time to figure out what's up with that right now - are there other packages on which you successfully ran grayskull with this PR?

I'm not even sure why `grayskull` would try and build the package - clearly that isn't going to work when there are external dependencies, unless it already parsed `[external]` and installed those deps with conda.

msarahan:
To be honest, I designed this only as proof of concept of reading the pyproject.toml tags and translating them into conda recipe equivalents. My tests were limited to unit tests of that translation process. Grayskull can be pointed at local sdists, though, so I will expand this to adopt your full integration testing with building packages from `external-deps-build`.

rgommers:
Thanks Mike, it'd be great to see it work on an sdist indeed.

I have only used `grayskull` a handful of times, so I may be missing something about what it's supposed to do or how to not make it try to build the package.

msarahan:
This goes much deeper than I thought. The translation happens fine, but grayskull's later parts don't pass it through. They go to setup.py/.cfg schema before really going to conda. It seems to me that we have to have a more direct path from pyproject.toml, and maintain the legacy path without pyproject.toml

@marcelotrevisani can you give me any context on why you chose to go from pyproject -> setup_py dict -> conda recipe? Was is just duplicating some stuff to have 

```
pyproject -> conda
setup_py -> conda
```

marcelotrevisani:
> This goes much deeper than I thought. The translation happens fine, but grayskull's later parts don't pass it through. They go to setup.py/.cfg schema before really going to conda. It seems to me that we have to have a more direct path from pyproject.toml, and maintain the legacy path without pyproject.toml
> 
> @marcelotrevisani can you give me any context on why you chose to go from pyproject -> setup_py dict -> conda recipe? Was is just avoiding duplicating some stuff to have
> 
> ```
> pyproject -> conda
> setup_py -> conda
> ```

it is because there are some cases where the pyproject.toml is present but the dependencies/metadata are not all specified there. Take ``pytest`` as example, there is a pyproject.toml there, however, the metadata is in a ``setup.cfg``

msarahan:
Thanks. That's a bummer, but I'm grateful for the quick response and pointer. I'll keep poking at it.

marcelotrevisani:
I have seen a few cases where people maintain both, also other cases where part of the metadata is in a pyproject.toml and part in a setup.cfg file. well, chaos 😆

msarahan:
Where's the poop emoji when you need it?

marcelotrevisani:
if you need any help from my side, please let me know. :)

rgommers:
> I have seen a few cases where people maintain both, also other cases where part of the metadata is in a pyproject.toml and part in a setup.cfg file.

You can derive this quite easily from the `pyproject.toml` content though. The one for `pytest` doesn't have a `[project]` table, so there are no dependencies there. However, if there is a `[project]` table and `dependencies` under it is not marked as dynamic, you can simply parse the `pyproject.toml` contents and be done with that.

There may be other content you need; the only one that's often dynamic is `version` (but you have it in the sdist already). Having the default path not execute a build would be really helpful I'd think, and more generic.

marcelotrevisani:
> > I have seen a few cases where people maintain both, also other cases where part of the metadata is in a pyproject.toml and part in a setup.cfg file.
> 
> You can derive this quite easily from the `pyproject.toml` content though. The one for `pytest` doesn't have a `[project]` table, so there are no dependencies there. However, if there is a `[project]` table and `dependencies` under it is not marked as dynamic, you can simply parse the `pyproject.toml` contents and be done with that.
> 
> There may be other content you need; the only one that's often dynamic is `version` (but you have it in the sdist already). Having the default path not execute a build would be really helpful I'd think, and more generic.

Yeah, that is something that I am working on a side project that I plan to integrate with grayskull with other bits, but it might take a while... :/

It is possible to add an option to give priority for pyproject.toml and get everything from it and not run the setup

msarahan:
I got a successful build with a skeleton generated from cffi. This stuff still needs polish before the PR is really ready. I have this code outputting a comment when a match can't be made for the PURL. I think that's nice, but I putting in comments is new in Grayskull, and it's glitchy. For example,

```
requirements:
  build:
    - "{{ compiler('c') }}"
  host:
    - python  #
# warning: no matching mapping was found for purl "pkg:generic/libffi". the dependency has been omitted here.
    - pip
```

Here the trailing comment after the python dependency shouldn't be here, and the indentation of the comment is wrong. he The indentation doesn't break the build, but it definitely doesn't look right.

Anyway, more to come.

grayskull/strategy/py_base.py
```diff
@@ -546,10 +547,12 @@ def clean_list_pkg(pkg, list_pkgs):
         return [p for p in list_pkgs if pkg != p.strip().split(" ", 1)[0]]
 
     for pkg in requirements["host"]:
-        pkg_name = RE_DEPS_NAME.match(pkg).group(0)
-        if pkg_name in PIN_PKG_COMPILER.keys():
-            requirements["run"] = clean_list_pkg(pkg_name, requirements["run"])
-            requirements["run"].append(PIN_PKG_COMPILER[pkg_name])
+        pkg_name_match = RE_DEPS_NAME.match(pkg)
```
msarahan, line 550: The reason for this change is that with PURLs, the deps name regex may not match. If that happens, then we are trying to call None.group(0) which of course fails.

msarahan:
@conda/grayskull I think this is ready for some review/discussion. Please check the original PR text for an updated summary of what's going on here.

marcelotrevisani (approved):

marcelotrevisani:
Sorry for my delay, the initial support looks good to me actually :)
Thanks for adding support for this

msarahan:
Thanks @marcelotrevisani. To clarify what comes next:

* The mapping for the PEP725 PURL names to conda package names needs to be created. @rgommers has a PEP draft that will follow PEP 725 in the branch at https://github.com/rgommers/peps/tree/pep-name-mapping. I believe the pypi<->conda-forge mapping is "county fair" in conda-forge, but I'm not sure if the PURL mapping will go there or in a new project.
* When the mapping project is available, this hard-coded stuff should be replaced: https://github.com/conda/grayskull/pull/518/files#diff-fb6b818bcb86e6ebb8733757d7942b0ca48568ee9b0f356de25c5e89fd4e1617R257
* In the interim, the hard-coded package mapping there can be expanded as necessary.
* I punted on handling optional dependencies, and this could be improved: https://github.com/conda/grayskull/pull/518/files#diff-fb6b818bcb86e6ebb8733757d7942b0ca48568ee9b0f356de25c5e89fd4e1617R284

If you see problems related to this new PEP725 stuff, please ping me. I'll be happy to help maintain this.

PR #541, opened 2024-07-12
Generate recipe from project tree using pypa/build

### LINKED ISSUE #542: [FEATURE] Use pypa/build to generate project metadata locally from any build system
Use pypa/build to get project metadata in a standard, build-system-independent way.

**Additional context**
Grayskull should be able to generate a recipe for a local project without referring to pypi, and without requiring a specific build system or pyproject.toml layout. Use pypa/build to generate the project's metadata, then shuffle that data into the recipe.

#541

dholth:
### Description

This is the start of something I've wanted: `grayskull pypi .` on a checkout to generate a recipe, suitable for committing to the project tree. Like https://github.com/conda/conda/blob/main/recipe/meta.yaml#L5-L7 where source is "../" or the parent directory of the meta.yaml.

pypa/build gets all project metadata in a standard way, including the build-system requires (setuptools, flit, poetry). If these are not already installed, we should add a "conda install <requirement>" before `project.metadata_path(...)` or bail and let the user install those. We would also change the pip command used in the build: section so that we never accidentally pip install a build requirement.

Remaining to be done, process all possible metadata out of `distribution` and into `recipe`; possibly deal with late "we don't know the package name until after `fetch_data`" more elegantly since this breaks grayskull's current assumptions.

It's also a little awkward that the recipe output dir is always `<given directory>/<project name>`. These would be `<checkout>/conda.recipe/meta.yaml` when included as first-party recipes.

Fix #542

### DIFF
--- a/grayskull/strategy/py_build.py
+++ b/grayskull/strategy/py_build.py
@@ -8,13 +8,15 @@ import tempfile
 from importlib.metadata import PathDistribution
 from pathlib import Path
 
-import build
-from packaging.markers import Marker
-from packaging.metadata import Metadata
+from conda.exceptions import InvalidMatchSpec
+from conda.models.match_spec import MatchSpec
+from packaging.requirements import Requirement
 from souschef.recipe import Recipe
 
+import build
 from grayskull.config import Configuration
 from grayskull.strategy.abstract_strategy import AbstractStrategy
+from grayskull.strategy.pypi import compose_test_section
 
 log = logging.getLogger(__name__)
 
@@ -24,16 +26,15 @@ class PyBuild(AbstractStrategy):
     def fetch_data(recipe: Recipe, config: Configuration, sections=None):
         project = build.ProjectBuilder(config.name)
 
+        recipe["source"]["path"] = "../"
+        # XXX relative to output. "git_url: ../" is also a good choice.
+
         with tempfile.TemporaryDirectory(prefix="grayskull") as output:
             build_system_requires = project.build_system_requires
             requires_for_build = project.get_requires_for_build("wheel")
             # If those are already installed, we can get the extras requirements
             # without invoking pip e.g. setuptools_scm[toml]
             print("Requires for build:", build_system_requires, requires_for_build)
-
-            recipe["requirements"]["host"] = sorted(
-                (*build_system_requires, *requires_for_build)
-            )
 
             # build the project's metadata "dist-info" directory
             metadata_path = Path(project.metadata_path(output_directory=output))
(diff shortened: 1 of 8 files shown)


grayskull/strategy/py_build.py
```diff
@@ -0,0 +1,144 @@
...
+
+            # grayskull thought the name was the path. correct that.
+            if recipe[0] == '#% set name = "." %}':  # XXX fragile
+                # recipe[0] = x does not work
+                recipe._yaml._yaml_get_pre_comment()[0].value = (
+                    f'#% set name = "{config.name}" %}}\n'
+                    f'#% set version = "{config.version}" %}}\n'
+                )
+
+                recipe["package"]["version"] = "{{ version }}"
```
dholth, line 56: This puts quotes around `version: '{{ version }}'`; how do we control the YAML more closely with conda-souschef's Recipe?

grayskull/strategy/py_build.py
```diff
@@ -0,0 +1,144 @@
...
+            recipe["requirements"]["host"] = [requires_python] + sorted(
+                (*build_system_requires, *requires_for_build)
+            )
+
+            requirements = [Requirement(r) for r in distribution.requires]
+            active_requirements = [
+                str(r).rsplit(";", 1)[0]
+                for r in requirements
+                if not r.marker or r.marker.evaluate()
+            ]
```
dholth, line 77: How do we include the marker as a YAML comment?

grayskull/strategy/py_build.py
```diff
@@ -0,0 +1,144 @@
...
+                f"{ep.name} = {ep.value}"
+                for ep in distribution.entry_points
+                if ep.group == "console_scripts"
+            ]
+
+            recipe["build"]["noarch"] = "python"
+            recipe["build"][
+                "script"
+            ] = "{{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation"
+            # XXX also --no-index?
```
dholth, line 106: It would be nice to be careful about not fetching pip requirements, even the build-system requirements (poetry, flit-core, hatchling, ...)

grayskull/strategy/py_build.py
```diff
@@ -0,0 +1,144 @@
...
+            metadata_dict = dict(
+                metadata
+            )  # XXX not what compose_test_section expects at all
+            metadata_dict["name"] = config.name
+            metadata_dict["entry_points"] = [
+                f"{ep.name} = {ep.value}"
+                for ep in distribution.entry_points
+                if ep.group == "console_scripts"
+            ]
+            recipe["test"] = compose_test_section(metadata_dict, [])
```
dholth, line 142: Maybe this sort of "copy all fields from one format to another" code is destined to be messy...

grayskull/utils.py
```diff
@@ -108,6 +108,12 @@ def origin_is_local_sdist(name: str) -> bool:
     )
 
 
+def origin_is_tree(name: str) -> bool:
+    """Return True if it is a directory"""
+    path = Path(name)
+    return path.is_dir()
```
dholth, line 114: We could look for pyproject.toml or setup.py; it should work with either.

dholth (commented):
Thanks for this useful project

grayskull/strategy/py_build.py
dholth: These are evaluated in the environment grayskull runs in, and might not be correct for the end user.

marcelotrevisani:
is this still as a draft?

dholth:
I would like feedback but it is still a draft, it's not clean yet.

grayskull/strategy/py_build.py
marcelotrevisani: conda-souschef is just a wrapper around ruamel yaml, you can still manipulate the ruamel yaml object directly.
you can do something like:
```python
recipe["package"]["version"].value = "{{ version }}"
```

grayskull/strategy/py_build.py
marcelotrevisani: there is a property called ``.comment`` that you can use for that
a recipe/section object uses this mixin
https://github.com/marcelotrevisani/souschef/blob/92cee9159fb65fa035593f1a8b60a931413ab581/src/souschef/mixins.py#L14

grayskull/strategy/py_build.py
marcelotrevisani: sorry, I think I didn't follow, what do you mean?

grayskull/utils.py
marcelotrevisani: that can work, but you also need to look for ``setup.cfg``

dholth: With the pypa/build method, pyproject.toml is the only necessary file to get the metadata.

grayskull/strategy/py_build.py
```diff
@@ -0,0 +1,158 @@
...
+            config.name = distribution.name  # see also _normalized_name
+            config.version = distribution.version
+
+            # grayskull thought the name was the path. correct that.
+            if recipe[0] == '#% set name = "." %}':  # XXX fragile
+                # recipe[0] = x does not work
+                recipe._yaml._yaml_get_pre_comment()[0].value = (
+                    f'#% set name = "{config.name}" %}}\n'
+                    f'#% set version = "{config.version}" %}}\n'
+                )
```
beenje, line 68: Instead of that, you can use

```python
set_global_jinja_var(recipe, "name", config.name)
set_global_jinja_var(recipe, "version", config.version)
```

grayskull/strategy/py_build.py
```diff
@@ -0,0 +1,158 @@
...
+"""
+
+import logging
+import tempfile
+from importlib.metadata import PathDistribution
+from pathlib import Path
+
+import build
+from conda.exceptions import InvalidMatchSpec
+from conda.models.match_spec import MatchSpec
```
beenje, line 13: Having to import conda will make grayskull more difficult to install.
Is it worth adding conda as dependency just for that function?

I haven't checked how requirements are parsed in the current code.

grayskull/strategy/py_build.py
```diff
@@ -0,0 +1,158 @@
...
+            # Description-Content-Type
+            # License-File
+            # License-File
+            # Requires-Dist (many times)
+            # Provides-Extra (several times)
+            # Description or distribution.metadata.get_payload()
+
+            about = {
+                "summary": metadata["summary"],
+                "license": metadata["license"],
```
beenje, line 141: On an example I tested, `license` was set to the full license text content. The `pyproject.toml` was using `license = {file = "LICENSE"}`. In the recipe, we want the SPDX identifier.

beenje (commented):
Very happy to see this PR!

I currently use `pyproject-build` to create a `sdist` and then run `grayskull` to generate the recipe from that local sdist.
This will make the process even easier.

One issue I stumbled upon is to find the module name to import (in the test section). It can differ from the distribution name.
I have a workaround when `setuptools` is used: https://github.com/conda/grayskull/pull/551

Do you know if this information can be retrieved from `build`?

dholth:
@beenje I have been [refining this technique here](https://github.com/dholth/conda-pupae/blob/main/conda_pupae/translate.py). It makes a wheel and immediately converts it to `.conda` without bothering with conda-build or a recipe. But it's focused on editable installs and there's no supporting code to manage build dependencies etc. In the future the metadata convert part might be shareable.

pyproject.toml doesn't give your imports in a generic way. You could inspect the contents of a wheel. You could have build-system-specific handling like how flit's name= must match the Python import.

grayskull/utils.py
marcelotrevisani: I agree, but not all projects uses it :/

beenje:
> @beenje I have been [refining this technique here](https://github.com/dholth/conda-pupae/blob/main/conda_pupae/translate.py). It makes a wheel and immediately converts it to `.conda` without bothering with conda-build or a recipe. But it's focused on editable installs and there's no supporting code to manage build dependencies etc. In the future the metadata convert part might be shareable.

Nice! Another project I have to look into :-)

grayskull/utils.py
dholth: pypa/build works with setup.py or pyproject.toml. It asks the underlying build system to create a wheel with `package.dist-info/METADATA`. Then we read `METADATA` and not `pyproject.toml`. So with this technique we don't touch setup.py, setup.cfg or pyproject.toml; pypa/build writes METADATA and we read that.

dholth:
https://conda.github.io/conda-pypi/

PR #543, opened 2024-07-19
fix license discovery: split at dashes for word-based matching

ytausch:
### Description

      
I noticed the following test is currently failing on main: `license/test_discovery/test_discovery.test_short_license_id` for `("3-Clause BSD License", "BSD-3-Clause")`.

Reason: There is a license called `DEC-3-Clause` which is as similar as `BSD-3-Clause` to `3-Clause BSD License` regarding the similarity measures we currently use. Splitting at dashes is a good way of handling cases correctly in which the license name is given without dashes.

It should be noted that the number of tests is very sparse so there is a possibility of breaking something else with this change. I still think it is safe to proceed since we are only modifying the behavior in situations we are unsure anyway.

    

### Additional Note
The other tests failing on main are not related to this PR.

### DIFF
--- a/grayskull/license/discovery.py
+++ b/grayskull/license/discovery.py
@@ -62,6 +62,16 @@ def get_all_licenses_from_spdx() -> List:
     ]
 
 
+def _replace_dashes(s: str) -> str:
+    """
+    Replace dashes with spaces.
+
+    :param s: string to replace dashes with spaces
+    :return: string with dashes replaced by spaces
+    """
+    return s.replace("-", " ")
+
+
 def match_license(name: str) -> dict:
     """Match if the given license name matches any license present on
     spdx.org
@@ -112,12 +122,21 @@ def match_license(name: str) -> dict:
                 lic[0] for lic in original_matches if lic[1] >= spdx_license[1]
             ]
             if len(best_matches) > 1:
+                # we replace dashes by spaces here to match instances like
+                # "3-Clause BSD" with "BSD-3-Clause" which otherwise would
+                # not work with word-based scores like token_sort_ratio
                 spdx_license = process.extractOne(
-                    name, best_matches, scorer=token_sort_ratio
+                    name,
+                    best_matches,
+                    scorer=token_sort_ratio,
+                    processor=_replace_dashes,
                 )
             if original_matches and original_matches[0][1] < 0.55:
                 spdx_license = process.extractOne(
-                    name, [m[0] for m in original_matches], scorer=token_sort_ratio
+                    name,
+                    [m[0] for m in original_matches],
+                    scorer=token_sort_ratio,
+                    processor=_replace_dashes,
                 )
 
     if spdx_license[1] != 100 and spdx_license[0].startswith("MIT"):


dholth:
I fixed the problem in a similar way https://github.com/conda/grayskull/pull/541/commits/9ba91d8f0b0f7a7222f1871a2e497ccb5b7206ca

ytausch:
The rapidfuzz tokenizer only splits at spaces, not at other characters. [This](https://github.com/rapidfuzz/RapidFuzz/blob/c80e7e14bd003b496b20803ea0b3880b763fa402/src/rapidfuzz/fuzz_py.py#L35-L37) is the relevant source code.

Regarding your fix: I think it confuses the following pairs of licenses:

```text
AFL-1.2 AFL-2.1
GPL-1.0 GPL-1.0+
GPL-2.0 GPL-2.0+
GPL-3.0 GPL-3.0+
LGPL-2.0 LGPL-2.0+
LGPL-2.1 LGPL-2.1+
LGPL-3.0 LGPL-3.0+
OLDAP-1.2 OLDAP-2.1
OLDAP-1.2 OLDAP-2.2.1
OLDAP-2.1 OLDAP-2.2.1
OLDAP-2.2 OLDAP-2.2.2
```

I obtained this list using a little Python script:

<details>

```python
import urllib.request
import json
import re

if __name__ == '__main__':
    with urllib.request.urlopen("https://github.com/spdx/license-list-data/raw/main/json/licenses.json") as url:
        data = json.load(url)
    for i, first_license in enumerate(data["licenses"]):
        first_id = first_license["licenseId"]
        first_bag = set(re.findall(r'\w+', first_id.lower()))
        for second_license in data["licenses"][i+1:]:
            second_id = second_license["licenseId"]
            second_bag = set(re.findall(r'\w+', second_id.lower()))
            if first_bag == second_bag:
                print(first_id, second_id)
```

</details>


Probably we should add a test like this as well.

ytausch:
I now ensured in an additional tests that license IDs are always mapped to themselves

tests/license/test_discovery.py
```diff
@@ -77,6 +77,14 @@ def test_short_license_id(licence_name, short_licence):
     assert get_short_license_id(licence_name) == short_licence
 
 
+@pytest.mark.parametrize(
+    "license_id", [lic["licenseId"] for lic in get_all_licenses_from_spdx()]
+)
+def test_short_license_id_map_to_self(license_id: str):
+    print(license_id)
```
marcelotrevisani, line 84:
```suggestion
```

marcelotrevisani: Can we remove this please?

ytausch: oops, yes of course

ytausch:
@marcelotrevisani I think the failing tests are not related to the changes I made in this PR

marcelotrevisani:
> @marcelotrevisani I think the failing tests are not related to the changes I made in this PR

yes, they are not related indeed, gonna take a look and I will try to fix them later.
I am not having much time these days to maintain the project due to work and personal life, but I am trying to keep a look into this :)
Thank you for your contribution!

marcelotrevisani:
can you rebase this branch to main please?

ytausch:
Hmm, not sure which parts of this PR are still relevant since @beenje merged #552 which is also targeted at the DEC/BSD confusion.

PR #548, opened 2024-08-13
Drop Jinja `name` from PyPI URL (using f-string instead)

jakirkham:
### Description

Instead of using Jinja to add `name` into the PyPI URL. Just use an f-string to template the URL. This removes `name` from the PyPI URL entirely and lets users more easily change this as needed.

Alternative to PR: https://github.com/conda/grayskull/pull/460


jakirkham:
It seems like it should be possible to drop this logic as well

https://github.com/jakirkham/grayskull/blob/d23e5ce7f985bec9b17a71e30299466048f50ef2/grayskull/strategy/pypi.py#L512-L520

Though please let me know if I'm missing something

grayskull/strategy/pypi.py
```diff
@@ -288,7 +288,7 @@ def get_pypi_metadata(config: Configuration) -> dict:
         "url": info.get("home_page"),
         "license": info.get("license"),
         "source": {
-            "url": config.url_pypi + "/packages/source/{{ name[0] }}/{{ name }}/"
+            "url": config.url_pypi + f"/packages/source/{config.name[0]}/{config.name}/"
```
bollwyvl, line 291: These will always be normalized (even for historic builds), with the most-widely used implementation being [`packaging.utils.canonicalize_name`](https://packaging.pypa.io/en/stable/utils.html#packaging.utils.canonicalize_name).

marcelotrevisani: that is a good idea, grayskull still tries to canonicalize the names, but having that implementation to use out of the box might be better indeed

jakirkham: Should this be stored as a key for reuse?

mgorny:
Oh, i See #598 is doing the same thing.

PR #539, opened 2024-05-16
Starts work on adding recipe upgrade feature

schuylermartin45:
### Description
First-pass implementation supporting the V1 recipe standard (used by `rattler-build`, as discussed in #486 

~~As of writing this PR (currently in the `DRAFT` stage), there are a few issues:~~
~~1. `conda-recipe-manager` is not currently in the conda-forge channel. There is an [open PR](https://github.com/conda-forge/staged-recipes/pull/26371) waiting for a review to add it in.
2. Locally, I can't get `ruamel` to dump `souschef`'s version of the recipe data into a `StringIO` stream. I had many issues getting conda to use my local copy of `conda-recipe-manager` and I _think_ it's related to that. I have `ruamel.yaml.jinja` installed in the local environment, but I still get an error. Here are some logs for more details:~~

```sh
(grayskull) smartin@Schuylers-MacBook-Pro:grayskull> grayskull pypi -u types-toml                       



#### Initializing recipe for types-toml (pypi) ####

Recovering metadata from pypi...
Starting the download of the sdist package types-toml
types-toml 100% Time:  0:00:00   1.8 MiB/s|####################################################################################|
Checking for pyproject.toml
pyproject.toml not found.
Recovering information from setup.py
Executing injected distutils...
Recovering metadata from setup.cfg
No data was recovered from setup.py. Forcing to execute the setup.py as script
Recovering metadata from setup.cfg
Checking >> -- 100% |#                                                                                  |[Elapsed Time: 0:00:00]
Recovering license info from spdx.org ...
INFO:Best match for license Apache-2.0 was Apache-2.0.
Best matches: [('Apache-2.0', 100.0, 0), ('Apache-1.0', 90.0, 1), ('Apache-1.1', 80.0, 2), ('Apache License 1.0', 67.5, 4), ('Zend-2.0', 55.55555555555556, 3)]
INFO:Github url: https://api.github.com/repos/python/typeshed/license?ref=0.10.8.20240310 - recovering license info
Recovering license information from github...
Recovering license info from repository...
Cloning into '/var/folders/mj/y6v4_y555yx3_370k4sk0_bh0000gp/T/gs-clone-repo-amc55x71'...
fatal: Remote branch 0.10.8.20240310 not found in upstream origin
Recovering license info from repository...
Cloning into '/var/folders/mj/y6v4_y555yx3_370k4sk0_bh0000gp/T/gs-clone-repo-a_k02yuu'...
[...]

PR #550, opened 2024-08-16
Fix meta.yaml output directory

beenje:
### Description

A small bug was introduced when [adding v1 format](https://github.com/conda/grayskull/pull/539).
The `meta.yaml` file was always created in the current directory instead of `recipe_dir`.


beenje:
Closing in favour of https://github.com/conda/grayskull/pull/552

PR #552, opened 2024-08-17
Fix tests

beenje:
### Description

Some tests are currently broken on `main`. Trying to fix those:

- fix `test_short_license_id` (`DEC-3-Clause' detected instead of `BSD-3-Clause`) - thanks to @dholth (commit from https://github.com/conda/grayskull/pull/541)
- fix --pypi-url: always raised a RuntimeError (`args.url_pypi_metadata` has a default value and will never be empty)
- fix meta.yaml output directory (the meta.yaml file was always created in the current directory instead of `recipe_dir`)
- fix `test_poetry_langchain_snapshot`: replace pypi.io with pypi.org (pypi.org is the official domain)
- skip `test_console_script_toml_format` for python 3.12 right now. We need to find another package for that test. Do you know any?
- fix `test_merge_pypi_sdist_metadata` for python 3.12: bump gsw version
- fix `test_injection_distutils_compiler_gsw` for python 3.12: bump gsw version - note that packages can't be retrieved from pyproject.toml right now
- fix `test_ciso_recipe` for python 3.12: bump ciso version

### DIFF
--- a/tests/test_pypi.py
+++ b/tests/test_pypi.py
@@ -1308,6 +1308,11 @@ def test_notice_file_different_licence():
     assert recipe["about"]["license"] in ["MIT AND Apache-2.0", "Apache-2.0 AND MIT"]
 
 
+# Need to find another package for this test
+@pytest.mark.skipif(
+    sys.version_info >= (3, 12),
+    reason="consolemd setup.py requires lower than python 3.12",
+)
 def test_console_script_toml_format():
     recipe, _ = create_python_recipe("consolemd", version="0.5.1")
     assert recipe["build"]["entry_points"] == ["consolemd = consolemd.cli:cli"]
--- a/tests/test_pypi.py
+++ b/tests/test_pypi.py
@@ -663,12 +663,23 @@ def test_injection_distutils_setup_reqs_ensure_list():
 
 
 def test_merge_pypi_sdist_metadata():
-    config = Configuration(name="gsw", version="3.3.1")
+    config = Configuration(name="gsw", version="3.6.19")
     pypi_metadata = get_pypi_metadata(config)
     sdist_metadata = get_sdist_metadata(pypi_metadata["sdist_url"], config)
     merged_data = merge_pypi_sdist_metadata(pypi_metadata, sdist_metadata, config)
     assert merged_data["compilers"] == ["c"]
-    assert sorted(merged_data["setup_requires"]) == sorted(["numpy"])
+    assert sorted(merged_data["setup_requires"]) == sorted(
+        [
+            "build",
+            'numpy<3,>=2.0.0rc1; python_version >= "3.9"',
+            'oldest-supported-numpy; python_version < "3.9"',
+            "pip>9.0.1",
+            "setuptools>=42",
+            "setuptools_scm[toml]>=3.4",
+            "wheel",
+            "python >=3.8",
+        ]
+    )
 
 
 def test_update_requirements_with_pin():
--- a/tests/test_pypi.py
+++ b/tests/test_pypi.py
@@ -179,7 +179,7 @@ def test_get_extra_from_requires_dist():
 def dask_sdist_metadata():
     config = Configuration(name="dask")
     return get_sdist_metadata(
-        "https://pypi.io/packages/source/d/dask/dask-2022.6.1.tar.gz",
+        "https://pypi.org/packages/source/d/dask/dask-2022.6.1.tar.gz",
         config,
     )
 
@@ -283,7 +283,7 @@ def test_compose_test_section_with_console_scripts():
     config = Configuration(name="pytest", version="7.1.2")
     metadata1 = get_pypi_metadata(config)
     metadata2 = get_sdist_metadata(
-        "https://pypi.io/packages/source/p/pytest/pytest-7.1.2.tar.gz", config
+        "https://pypi.org/packages/source/p/pytest/pytest-7.1.2.tar.gz", config
     )
     metadata = merge_pypi_sdist_metadata(metadata1, metadata2, config)
     test_requirements = []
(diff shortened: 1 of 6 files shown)


beenje:
`test_injection_distutils_compiler_gsw` is failing on python 3.12 due to:
`DEBUG    grayskull.strategy.py_base:py_base.py:460 Exception when executing setup.py as script: module 'configparser' has no attribute 'SafeConfigParser'`

`SafeConfigParser` was removed in 3.12.
This comes from `versioneer.py` included in the sdist. Test should be updated to use another package/version.

tests/test_pypi.py
```diff
@@ -809,13 +820,22 @@ def test_download_pkg_sdist(pkg_pytest):
...
     recipe = GrayskullFactory.create_recipe(
-        "pypi", Configuration(name="ciso", version="0.1.0")
+        "pypi", Configuration(name="ciso", version="0.2.2")
     )
     assert sorted(recipe["requirements"]["host"]) == sorted(
-        ["cython", "numpy", "pip", "python"]
+        [
+            "cython >=3",
+            "numpy >=2.0.0rc1",
+            "oldest-supported-numpy",
```
beenje, line 829: We shouldn't have both `numpy >=2.0.0rc1` and `oldest-supported-numpy` but it looks like grayskull doesn't support selectors?

marcelotrevisani: which selectors are you talking about? It does support a few of them but in this case they might not be being generated for some reason

marcelotrevisani:
thanks for fixing the tests, much appreciated! :)

marcelotrevisani (approved):

tests/test_pypi.py
jakirkham: Yeah this is `requirements/host`. Don't think we want `oldest-supported-numpy` here

tests/test_pypi.py
```diff
@@ -809,13 +820,22 @@ def test_download_pkg_sdist(pkg_pytest):
...
+            "pip",
+            "python >=3.9",
+            "setuptools >=41.2",
+            "setuptools-scm",
+            "wheel",
+        ]
     )
     assert sorted(recipe["requirements"]["run"]) == sorted(
-        ["cython", "python", "<{ pin_compatible('numpy') }}"]
+        ["<{ pin_compatible('numpy') }}", "oldest-supported-numpy", "python >=3.9"]
```
jakirkham, line 838: Also not sure why `oldest-supported-numpy` is added to `requirements/run`

marcelotrevisani: Hi @jakirkham , at first sight it seems odd, indeed, but it comes directly from the requirements in that package ``gsw``, it is not something that grayskull adds, 
https://github.com/TEOS-10/GSW-Python/blob/c3407368ef01d0b3d8e85f8cf7b94abae52bcd25/pyproject.toml#L6

that is a separated package on pypi
https://pypi.org/project/oldest-supported-numpy/

tests/test_pypi.py
marcelotrevisani: see my comment here ;)
https://github.com/conda/grayskull/pull/552/files#r1721631804

marcelotrevisani: that equal over there is just a syntax sugar, but if you properly compare the ``numpy`` and ``oldest-supported-numpy`` you will see the selectors :)

beenje: Ah! Indeed, looking at the generated meta.yaml, I see:

```
    - oldest-supported-numpy  # [py<39]
    - numpy >=2.0.0rc1  # [py>=39]
 ```

File: tests/test_pypi.py (hunk 2 above)
Comment (beenje) [reply tests/test_pypi.py:838]: Indeed this is actually for the [ciso](https://github.com/ioos/ciso) project that has `oldest-supported-numpy ` not only in build-system requirements , but in [dependencies]( https://github.com/ioos/ciso/blob/189ad38f84bc20176ad46b312fd312eddd548c53/pyproject.toml#L23C1-L26C2) as well:

```
dependencies = [
    "oldest-supported-numpy ; python_version < '3.9'",
    "numpy>=2.0.0rc1 ; python_version >= '3.9'",
]
```

jakirkham: It's ok to have in `pyproject.toml`

However what this appears to be testing is whether this is added to `requirements/host` and `requirements/run`. The latter doesn't make sense

marcelotrevisani: it is going to appear in the requirements run because ``oldest-supported-numpy`` is specified in the dependencies section of the pyproject.toml. If it is was only in build section of pyproject.toml that would appear only in host then.

It is possible to do a workaround and always remove it from the requirements/run and always let it in host, but I am just not so sure if it covers all scenarios, what do you think?

jakirkham: We should remove `oldest-supported-numpy` whenever it is seen. That's used by wheels specifically to configure builds (not sure why a wheel would use `oldest-supported-numpy` at runtime unless they are some kind of build tool)

In any event it doesn't make sense in the Conda case where we configure NumPy versions used in some `conda_build_config.yaml` or other [variants]( https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html )

On the Conda side, we need only add `numpy` (unconstrained) to `requirements/host` and copy over any `dependencies` value of `numpy` to `requirements/run`

marcelotrevisani: ok, gotcha
That is pretty much a new PR to do this workaround as it was not a requirement before.

jakirkham: Gotcha thanks for the context! 🙏

The issue is `oldest-supported-numpy` doesn't exist in Conda. So if Grayskull adds it to recipes, they will be broken. Maintainers will need to know they can clean this out to fix the issue

In any event, we can discuss more in the new issue (thanks for raising! 🙏): https://github.com/conda/grayskull/issues/553

PR #557, opened 2024-09-06
Split parsing of Poetry versions into a separate module

maresb:
### Description

While working on conda-lock I noticed that our parsing of pyproject.toml files is incorrect. I'd like to copy it from here. In preparation for this I cleaned up some stuff and split it into a separate module.

When splitting the files I did some fancy Git tricks to preserve the blame history. (That's what the "Split history ..." commits are about.)

Also, as a personal preference, for unit tests regarding the nominal operation of functions I really like doctests. I enabled them. Of course if you don't like that I can revert those commits.

While this PR as a whole may be somewhat confusing, I really try to make each individual commit really easy and quick to understand and verify. I hope I achieved that, but if not, please ask. Thanks a lot for your time!


maresb:
Ok, the only failing tests now are `test_short_license_id`, which I'm pretty sure is independent of my changes here.

maresb:
Please have a look at this particular commit: https://github.com/maresb/grayskull/commit/d367c1605f320bae345b441b81ba38f5f72dc17e

There is an extra whitespace character before `sampleSelection`, and `rgl` is missing from the imports, and there is an extra empty string. I don't know enough about R to know whether or not these changes to the expected output are consequential, but since I'm making a real doctest out of them, I want to get an explicit signoff.

.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     - name: Running serial tests
```
marcelotrevisani, line 38: can you revert the modifications in this file please?

marcelotrevisani:
that is a nice refactoring, thank you for that!
I just added a comment, if you could change it, it would be nice and after that we can merge it :)

.github/workflows/tests.yml
maresb: Cool, thanks so much for the review!

I can certainly rearrange the tests, but could you explain a bit about the current structure so that I can implement it correctly? In particular, should the doctests execute before or after the serial/parallel tests?

I don't understand why the tests are currently split into serial/parallel, and from what I can see they are executing in exactly the same context, so I'm confused. By default, unless there's some good reason, I'd expect the tests to run all together in the same job, so that if some of the tests fail then you see all the failing tests rather than only the failing tests in the particular job.

maresb:
I backed out of the merge commit and moved the "Enable doctests" commit to be redone to the end.

I will modify it and then redo the merge commit.

At this point there are no changes to the code, only shuffling around commits.

maresb:
@marcelotrevisani, I made an initial guess about what you want. No problem to change it if you'd like it done differently.

maresb:
My previous attempt to split the doctests was broken, but I fixed it and now the CI is green.

.github/workflows/tests.yml
marcelotrevisani: The tests are split into serial and parallel because the serial tests can interfere with other tests and make them fail.

marcelotrevisani:
Thanks a lot!

marcelotrevisani (approved):

maresb:
Wonderful, thank you @marcelotrevisani!!!

PR #558, opened 2024-09-12
Switch to Ruff tools, update pre-commit

mwtoews:
The aims of this PR are to switch to faster/modern Ruff-based lint and formatting tools, and to update the pre-commit workflow. Details:

- Replace "black", "isort" and "flake8" with [Ruff](https://docs.astral.sh/ruff/). Some of these rules are selected in pyproject.toml, and I think most others are enabled by default. But keep blacken-docs for now, with the black version updated.
- Remove unused configuration files (and references in MANIFEST.in)
- Remove [deprecated `fix-encoding-pragma`](https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#fix-encoding-pragma) which is replaced by pyupgrade/["UP"](https://docs.astral.sh/ruff/rules/#pyupgrade-up) Ruff rule. Other changes from this rule are removing default "r" from open() statements
- Update sort-simple-yaml to look for any "config.yaml" file (there wasn't any grayskull/pypi/config.yaml file). The change from this sorts keys in grayskull/strategy/config.yaml

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,26 +1,19 @@
 exclude: docs
 repos:
--   repo: https://github.com/psf/black
-    rev: 24.8.0
-    hooks:
-    -   id: black
-        args: [--safe, --quiet]
--   repo: https://github.com/asottile/blacken-docs
+-   repo: https://github.com/adamchainz/blacken-docs
     rev: 1.18.0
     hooks:
     -   id: blacken-docs
-        additional_dependencies: [black==22.6.0]
+        additional_dependencies: [black==22.12.0]
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.6.0
     hooks:
     -   id: trailing-whitespace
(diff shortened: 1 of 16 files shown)


.pre-commit-config.yaml
```diff
@@ -1,41 +1,31 @@
...
-    rev: 7.1.1
+-   repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.6.4
     hooks:
-    -   id: flake8
-        exclude: tests/data
-        language_version: python3
+      # Run the linter
+      - id: ruff
+        args: [ --fix ]
```
marcelotrevisani, line 29:
```suggestion
        args: [ --fix, --exit-non-zero-on-fix ]
```

marcelotrevisani:
Thanks for that, I wanted to move to use ruff but I was always forgetting about it.

marcelotrevisani (approved):

PR #562, opened 2024-09-27
Require Python 3.11

maresb:
### Description

The CI is only testing Python 3.11, and the Conda feedstock requires Python 3.11 and 3.12, so I assume that we only intend to support >=3.11. This reflects this intention in the PyPI metadata.


maresb:
Probably the better way to go here is #563

maresb:
Closing in favor of #563

PR #563, opened 2024-09-27
Test on more versions of Python

maresb:
### Description
Dual to #562 let's check which Python versions work.

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -29,4 +29,3 @@ dependencies:
   - libcblas
   - beautifulsoup4
   - semver >=3.0.0,<4.0.0
-  - conda-recipe-manager >=0.2.0
--- a/grayskull/utils.py
+++ b/grayskull/utils.py
@@ -244,7 +244,15 @@ def upgrade_v0_recipe_to_v1(recipe_path: Path) -> None:
           JINJA plugin.
     :param recipe_path: Path to that contains the original recipe file to modify.
     """
-    from conda_recipe_manager.parser.recipe_parser_convert import RecipeParserConvert
+    try:
+        from conda_recipe_manager.parser.recipe_parser_convert import (
+            RecipeParserConvert,
+        )
+    except ImportError as e:
+        raise ImportError(
+            "Please install conda-recipe-manager from conda-forge to enable "
+            "support for the V1 format. (Note that Python >=3.11 is required.)"
+        ) from e
 
     recipe_content: Final[str] = RecipeParserConvert.pre_process_recipe_text(
         recipe_path.read_text()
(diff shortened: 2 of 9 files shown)


maresb:
Ok, conda-recipe-manager fails unless Python >=3.11

maresb:
Probably the right thing to do is conditionally install conda-recipe-manager on Python >=3.11 and then conditionally run only the tests that actually need it.

marcelotrevisani:
yeah, that is due to the fact that ``conda-recipe-manager`` constrain python
but that is a optional dependency, we can indeed run for older python versions without it

maresb:
It looks like there is no coverage yet for `--use-v1-format` so the tests already pass without `conda-recipe-manager`.

maresb:
Oops, seems to be hitting https://github.com/mamba-org/setup-micromamba/issues/225

maresb:
@marcelotrevisani, what do you think of these changes?

Tests are now passing down to Python 3.8. In order to get there I had to `from __future__ import annotations` to some files, tweak a context manager call, and run `ruff check --fix`.

I removed `conda-recipe-manager` from the `environment.yml`, but in the CI I conditionally install it when the Python version is >=3.11.

Finally, I switched from `setup-miniconda` to `setup-micromamba`.

environment.yaml
```diff
@@ -29,4 +29,4 @@ dependencies:
   - libcblas
   - beautifulsoup4
   - semver >=3.0.0,<4.0.0
-  - conda-recipe-manager >=0.2.0
+  - micromamba
```
marcelotrevisani, line 32: do we need micromamba here? 😬

grayskull/utils.py
```diff
@@ -244,7 +244,15 @@ def upgrade_v0_recipe_to_v1(recipe_path: Path) -> None:
...
-    from conda_recipe_manager.parser.recipe_parser_convert import RecipeParserConvert
+    try:
+        from conda_recipe_manager.parser.recipe_parser_convert import (
+            RecipeParserConvert,
+        )
+    except ImportError as e:
+        raise ImportError(
+            "Please install conda-recipe-manager from conda-forge to enable "
+            "support for the V1 format. (Note that Python >=3.11 is required.)"
+        ) from e
```
marcelotrevisani, line 255: cool, thanks for that

marcelotrevisani:
hey, thanks again for this PR, I just added one comment about micromamba in the environment.yaml, do we need it there?

maresb:
Ah, very good point!!! Now that I switched the CI over to `setup-micromamba` I don't think we need it since I needed it specifically for that CI.

I just rebased to remove the commit where I added it. (See the rebase diff with the "Compare" button or equivalently [here](https://github.com/conda/grayskull/compare/a078d44870a0d6ea688972ff3cbe2bdacda823b0..137d60c4c172ad40e74eb7610c9be9c37c8f12be).)

maresb:
@marcelotrevisani, yep, it's working. The relevant part of CI has completed.

marcelotrevisani (approved):

PR #573, opened 2024-11-06
Handle or operator in Poetry dependency specification

maresb:
### Description

It's not documented in Poetry's [dependency specification](https://python-poetry.org/docs/dependency-specification/) but they support an [or operator](https://github.com/python-poetry/poetry-core/blob/2cbe240beda1bb20eaf02a1ca0353b0ce6c167e6/src/poetry/core/constraints/generic/parser.py#L35).

I ran into this in the wild [here](https://github.com/jerry-git/pytest-split/blob/9cfb51f3bfb33fb10ee344f30292aea0f2705852/pyproject.toml#L37).


marcelotrevisani:
thank you again!

PR #584, opened 2024-11-18
MNT: Remove 'pip install' options in recipes

### LINKED ISSUE #582: Question: why is --no-deps --no-build-isolation set?
Conda-build sets `PIP_NO_BUILD_ISOLATION` and `PIP_NO_DEPENDENCIES` (See https://github.com/conda/conda-build/blob/0ad4bb6829b440caae9ae8a4db14ab1ff3788ab7/conda_build/build.py#L3020). Why is it also being set in every recipe in https://github.com/conda/grayskull/blob/5b293078ea4fb17572dfa63ee1dde8612d884f00/grayskull/strategy/pypi.py#L67 ? There are other flags set by conda-build, so this seem like an arbitrary set to re-specify on the command line.

I think both[^1] of these should be removed to allow conda-build to specify everything needed for pip to work and keep the recipes as simple as possible. Or all the other flags should be added too.

Notice in https://github.com/conda-forge/packaging-feedstock/pull/32 where `--no-build-isolation` was being added to a recipe based on greyskull.

[^1]: Not completely sure PIP_NO_DEPENDENCIES is the same as `--no-deps`, as it seems to be it should be `PIP_NO_DEPS`. The translation is automatic, and there isn't a `--no-dependencies` flag. But I am sure that `PIP_NO_BUILD_ISOLATION` is correct. So at least that one could be trimmed here. Edit: tested and PIP_NO_DEPENDENCIES does work

matthewfeickert:
Resolves #582

### Description

* Remove `--no-deps --no-build-isolation` `pip install` options in recipes as the build tool (e.g. `conda-build`, `rattler-build`) will enforce all required `pip install` options itself at build time.

c.f.:
* https://github.com/conda/conda-build/pull/5541
* https://github.com/prefix-dev/rattler-build/pull/1198

### DIFF
--- a/grayskull/strategy/pypi.py
+++ b/grayskull/strategy/pypi.py
@@ -64,9 +64,7 @@ class PypiStrategy(AbstractStrategy):
     def fetch_data(recipe, config, sections=None):
         update_recipe(recipe, config, sections or ALL_SECTIONS)
         if not (recipe["build"] and recipe["build"]["script"]):
-            recipe["build"]["script"] = (
-                "<{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation"
-            )
+            recipe["build"]["script"] = "<{ PYTHON }} -m pip install . -vv"
 
 
 def merge_pypi_sdist_metadata(
--- a/tests/data/poetry/langchain-expected.yaml
+++ b/tests/data/poetry/langchain-expected.yaml
@@ -12,7 +12,7 @@ build:
   entry_points:
     - langchain-server = langchain.server:main
   noarch: python
-  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
 requirements:


matthewfeickert:
This is ready for review.

ocefpaf:
@wolfv we can't do this while we use conda-build b/c of the multiple outputs bug. However, if/when we drop conda-build, do you know if rattler-build will behave as expected with multiple outputs?

matthewfeickert:
> we can't do this while we use conda-build b/c of the multiple outputs bug.

Out of curiosity, is there an issue for this?

ocefpaf:
> Out of curiosity, is there an issue for this?

I'm pretty sure there is one, or more, in conda-build. To be fair, are trying to keep conda-build on life support as we move to rattler-build, so I don't think anyone is working on that.

marcelotrevisani:
> @wolfv we can't do this while we use conda-build b/c of the multiple outputs bug. However, if/when we drop conda-build, do you know if rattler-build will behave as expected with multiple outputs?

after we drop conda-build and start to use rattle, we could add this when the ``--strict-conda-forge`` flag is enabled

matthewfeickert:
As I don't know the full history of things, or even the technical problems with `conda-build`, should I instead change this PR to **add** in `--ignore-installed` as that's also a flag set (https://github.com/conda/conda-build/issues/5543) and then once conda-forge has gone full `rattler-build` removing all the flags can be revisited?

I also don't know if there is a projected timeline for full rattler-build adoption. (2025? 2026?)

marcelotrevisani:
> As I don't know the full history of things, or even the technical problems with `conda-build`, should I instead change this PR to **add** in `--ignore-installed` as that's also a flag set ([conda/conda-build#5543](https://github.com/conda/conda-build/issues/5543)) and then once conda-forge has gone full `rattler-build` removing all the flags can be revisited?
> 
> I also don't know if there is a projected timeline for full `rattler-build` adoption. (2025? 2026?)

yeah, I think that would be nice to add the flag for now :)

matthewfeickert:
> ... then once conda-forge has gone full rattler-build removing all the flags can be revisited?

@marcelotrevisani conda-forge has now changed to use the v1 / `rattler-build` recipe format as the default (c.f. PR https://github.com/conda/grayskull/pull/669). @wolfv can correct me if I'm wrong, but I believe that the issue with the flags being needed for recipes with multiple build outputs is now also resolved (c.f. https://github.com/conda-forge/conda-forge.github.io/pull/2913, https://github.com/conda/grayskull/issues/582#issuecomment-5249292225, and https://github.com/conda/grayskull/issues/582#issuecomment-5262389300). So I think this PR is ready for review again.

matthewfeickert:
:wave: @marcelotrevisani @jezdez Can you give this a review in the next week?

matthewfeickert:
👋 @marcelotrevisani @jezdez ping on this.

matthewfeickert:
https://github.com/prefix-dev/rattler-build/pull/1198 has been merged now, removing the use of `--no-deps --no-build-isolation` in recipe recommendations.

matthewfeickert:
cc @jaimergp and @ocefpaf for review as well

marcelotrevisani (approved):

matthewfeickert:
Thanks @marcelotrevisani!

Issue #463, opened 2023-04-29
[BUG] lists in pyproject.toml dependencies not handled


xylar:
**Describe the bug**
Dependencies like the following in a `pyproject.toml` raise a `grayskull.strategy.py_toml.InvalidPoetryDependency` error:
```
pyarrow = [
    {version = ">=6.0.0", python = ">=3.7,<3.11"},
    {version = ">=10.0.1", python = ">=3.11"}
]
```

**To Reproduce**
```
$ grayskull pypi databricks-sql-connector

#### Initializing recipe for databricks-sql-connector (pypi) ####

Recovering metadata from pypi...
Starting the download of the sdist package databricks-sql-connector
databricks-sql-connector 100% Time:  0:00:00 664.7 KiB/s|#####################|
Checking for pyproject.toml
pyproject.toml found in /tmp/grayskull-databricks-sql-connector-tdnkdt4v/databricks_sql_connector-2.5.1/pyproject.toml
Traceback (most recent call last):
  File "/home/xylar/mambaforge/bin/grayskull", line 10, in <module>
    sys.exit(main())
  File "/home/xylar/mambaforge/lib/python3.10/site-packages/grayskull/__main__.py", line 276, in main
    generate_recipes_from_list(args.pypi_packages, args)
  File "/home/xylar/mambaforge/lib/python3.10/site-packages/grayskull/__main__.py", line 299, in generate_recipes_from_list
    recipe, config = create_python_recipe(
  File "/home/xylar/mambaforge/lib/python3.10/site-packages/grayskull/__main__.py", line 334, in create_python_recipe
    GrayskullFactory.create_recipe(
  File "/home/xylar/mambaforge/lib/python3.10/site-packages/grayskull/base/factory.py", line 46, in create_recipe
    GrayskullFactory.REGISTERED_STRATEGY[repo_type.lower()].fetch_data(
  File "/home/xylar/mambaforge/lib/python3.10/site-packages/grayskull/strategy/pypi.py", line 60, in fetch_data
    update_recipe(recipe, config, sections or ALL_SECTIONS)
  File "/home/xylar/mambaforge/lib/python3.10/site-packages/grayskull/strategy/pypi.py", line 485, in update_recipe
    metadata = get_metadata(recipe, config)
  File "/home/xylar/mambaforge/lib/python3.10/site-packages/grayskull/strategy/pypi.py", line 347, in get_metadata
    sdist_metadata, pypi_metadata = get_origin_wise_metadata(config)
[...]

Issue #535, opened 2024-04-15
[BUG] Failed to parse `tool.poetry.dependencies` with multiple constraints dependencies


t20100:
**Describe the bug**

`grayskull` fails to parse poetry's [multiple constraints dependencies](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies) syntax when run on the [xypattern](https://github.com/CPrescher/xypattern) package with a `grayskull.strategy.py_toml.InvalidPoetryDependency` exception (see below).

This is the [xypatterns's pyproject.toml `[tool.poetry.dependencies]` section](https://github.com/CPrescher/xypattern/blob/5ed6515c0ddba99997e4a188aef138efde3e5905/pyproject.toml#L37C2-L46).

**To Reproduce**

- Run `grayskull pypi xypattern`

**Expected behavior**

Have `grayskull` supporting poetry's [multiple constraints dependencies](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies) syntax.

**Outputs**

<details><summary>Raised exception</summary>

```
Checking for pyproject.toml
pyproject.toml found in /tmp/grayskull-xypattern-r04377da/xypattern-1.1.2/pyproject.toml
Traceback (most recent call last):
  File "py312/bin/grayskull", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "py312/lib/python3.12/site-packages/grayskull/main.py", line 277, in main
    generate_recipes_from_list(args.pypi_packages, args)
  File "py312/lib/python3.12/site-packages/grayskull/main.py", line 300, in generate_recipes_from_list
    recipe, config = create_python_recipe(
                     ^^^^^^^^^^^^^^^^^^^^^
  File "py312/lib/python3.12/site-packages/grayskull/main.py", line 335, in create_python_recipe
    GrayskullFactory.create_recipe(
  File "py312/lib/python3.12/site-packages/grayskull/base/factory.py", line 46, in create_recipe
    GrayskullFactory.REGISTERED_STRATEGY[repo_type.lower()].fetch_data(
  File "py312/lib/python3.12/site-packages/grayskull/strategy/pypi.py", line 62, in fetch_data
    update_recipe(recipe, config, sections or ALL_SECTIONS)
  File "lib/python3.12/site-packages/grayskull/strategy/pypi.py", line 499, in update_recipe
    metadata = get_metadata(recipe, config)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #523, opened 2024-01-15
Support Python-version-specific deps in poetry

xhochy:
Poetry also supports dependencies depending on the Python version. I have encountered this in the `nannyml` package.

PR #592, opened 2024-12-15
Poetry deps w/ python version and platform, and multiple constraints dependencies

lorepirri:
### Description

When reading [Poetry dependencies](https://python-poetry.org/docs/dependency-specification/) from a `pyproject.toml`, Grayskull does not read the requirements of `python` version and `platform`.

Also, it does not support [multiple constraints dependencies](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies) and it crashes (see #463 #535 and an open PR #523 by @xhochy ).

This PR aims to add some support for these two missing features.

As mentioned in the two issues, one could see `grayskull` crashing on:

- `grayskull pypi databricks-sql-connector`
- `grayskull pypi xypattern`
- `grayskull pypi ersilia`
- `grayskull pypi nannyml` (this one is still failing for a dependency that uses caret `^` and non `semver` standard version `^0.8.post1` and `grayskull` crashes when trying to coerce it to `semver` for the `floor`. See #593 )

Tests are added, including some basic regression tests for `databricks-sql-connector`, `xypattern`, `ersilia`, and `nannyml`.

Tagging also for visibility people active in the issues: @xylar @darynwhite @kcpevey @DhanshreeA @thewchan 

#### Platform

From [Poetry dependencies](https://python-poetry.org/docs/dependency-specification/) documentation, the requirement `platform` seems to have only two values `linux` and `darwin`, probably `windows` but not sure (respectively rendered as `linux`, `osx`, and `win` selectors).

It is unclear if one could use `platform = "!=darwin"` to express `  # [not osx]`.

#### Python version

`python` version is rendered naively and differently than for `pypi.py` and `py_base.py` (I did not get the function `generic_py_ver_to` with `py2k` and`py3k`, because a lot going on in that function and I am not familiar with those selectors). If the version is `>=3.10` this will simply result in a selector `py>=310`.

If this is not wanted or bloating the code, I am willing to re-use functions from `py_base.py` or `pypi.py` with the right guidance (thanks).

It does expand caret and tilde (e.g. `^3.10` -> `py>=310 and py<4`), and always removes `patch`, and also `minor` if `0` (e.g. `3.11.1` -> `py==311`, `3.12.0` -> `py==312`, and `4.0.0` -> `py==4`).

Multiple version specifiers are at the moment always rendered in `and`. For example `>=3.10.0,<3.11.0` with render `py>=310 and py<311`. Same as `conda-build` is doing. Ref: https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications

If one would want to express `<3.7` or `>=3.10` should use the pipe operator `<3.7|>=3.10`. In fact, `<3.7|>=3.10` renders `py<37 or py>=310` and `<3.8|>=3.10,!=3.11` renders `py<38 or py>=310 and py!=311` (`,` and `and` take precedence over `|` and `or`).

It does handle compatible release operator `~=` correctly and the wildcard `.*` suffix in the version as well (not very common though as the resulting selectors are not interesting).

#### Others

👉 Fixes a minor bug: the compatible release operator `~=` in the dependency version could not be used because the parser was jumping to expanding the tilde `~` and crashing on `=` (or returning the wrong version containing `=`). Therefore I suspect that the  `~=` operator was never handled (maybe never used, or nobody reported this bug).

What does not add:
- The requirement `markers`, `extras`, and others, are not supported (ignored). Nevertheless, the ground is set, and it is easier now to add support for this.
- [Expanded dependency specification syntax](https://python-poetry.org/docs/dependency-specification/#expanded-dependency-specification-syntax) not touched. I don't know if the current version of `grayskull` is able to handle it.

xref #463 #535 / PR #523

### DIFF
--- a/grayskull/strategy/parse_poetry_version.py
+++ b/grayskull/strategy/parse_poetry_version.py
@@ -260,9 +260,8 @@ def encode_poetry_python_version_to_selector_item(poetry_specifier: str) -> str:
     Example:
         ">=3.8,<3.12" => "py>=38 and py<312"
         ">=3.8,<3.12,!=3.11" => "py>=38 and py<312 and py!=311"
-        # TODO: how this case will render?
-        "<3.8,>=3.10" => "py<38 or py>=310"
-        "<3.8,>=3.10,!=3.11" => "(py<38 or py>=310) and py!=311"
+        "<3.8|>=3.10" => "py<38 or py>=310"
+        "<3.8|>=3.10,!=3.11" => "py<38 or py>=310 and py!=311"
 
     # handle exact version specifiers correctly
     >>> encode_poetry_python_version_to_selector_item("3")
@@ -284,15 +283,29 @@ def encode_poetry_python_version_to_selector_item(poetry_specifier: str) -> str:
     >>> encode_poetry_python_version_to_selector_item("~3.10") # '>=3.10.0,<3.11.0'
     'py>=310 and py<311'
 
-    # handle multiple requirements correctly
+    # handle multiple requirements correctly (in "and")
     >>> encode_poetry_python_version_to_selector_item(">=3.8,<3.12,!=3.11")
     'py>=38 and py<312 and py!=311'
+
+    # handle multiple requirements in "or" correctly ("and" takes precendence)
+    >>> encode_poetry_python_version_to_selector_item("<3.8|>=3.10,!=3.11")
+    'py<38 or py>=310 and py!=311'
     """
 
     if not poetry_specifier:
         return ""
 
     version_specifier = encode_poetry_version(poetry_specifier)
+
+    if "|" in version_specifier:
+        poetry_or_clauses = [clause.strip() for clause in version_specifier.split("|")]
+        conda_or_clauses = [
+            encode_poetry_python_version_to_selector_item(clause)
+            for clause in poetry_or_clauses
+            if clause != ""
+        ]
+        conda_or_clauses = " or ".join(conda_or_clauses)
+        return conda_or_clauses
 
     conda_clauses = version_specifier.split(",")
 
--- a/grayskull/strategy/parse_poetry_version.py
+++ b/grayskull/strategy/parse_poetry_version.py
@@ -260,6 +260,8 @@ def encode_poetry_python_version_to_selector_item(poetry_specifier: str) -> str:
     Example: ">=3.8,<3.12" => "py>=38 or py<312"
 
     # handle exact version specifiers correctly
+    >>> encode_poetry_python_version_to_selector_item("3")
+    'py==3'
     >>> encode_poetry_python_version_to_selector_item("3.8")
     "py==38"
     >>> encode_poetry_python_version_to_selector_item("==3.8")
@@ -299,17 +301,33 @@ def parse_python_version(selector: str):
     Return operator and normalized version from a version selector
 
     Examples:
+        ">=3"   -> ">=", "3"
+        ">=3.0"   -> ">=", "3"
         ">=3.8"   -> ">=", "3.8"
         ">=3.8.0" -> ">=", "3.8"
         "<4.0.0"  -> "<", "4"
         "3.12"    -> "==", 3.12"
         "=3.8"    -> "==", "3.8"
+        ">=3.8.0.1" -> ">=", "3.8"
+
+    >>> parse_python_version(">=3.8")
+    ('>=', '3.8')
+    >>> parse_python_version("3.12")
+    ('==', '3.12')
+    >>> parse_python_version("<4.0.0")
+    ('<', '4')
+    >>> parse_python_version(">=3")
+    ('>=', '3')
+    >>> parse_python_version(">=3.8.0")
+    ('>=', '3.8')
+    >>> parse_python_version(">=3.8.0.1")
+    ('>=', '3.8')
 
     The version is normalized to "major.minor" (drop patch if present)
     or "major" if minor is 0
     """
     # Regex to split operator and version
-    pattern = r"^(?P<operator>\^|~|>=|<=|!=|==|>|<|=)?(?P<version>\d+(\.\d+){0,2})$"
+    pattern = r"^(?P<operator>\^|~|>=|<=|!=|==|>|<|=)?(?P<version>\d+(\.\d+){0,3})$"
     match = re.match(pattern, selector)
     if not match:
         raise ValueError(f"Invalid version selector: {selector}")
@@ -321,7 +339,12 @@ def parse_python_version(selector: str):
     version = match.group("version")
 
     # Split into major, minor, and discard the rest (patch or additional parts)
-    major, minor, *_ = version.split(".")
+    try:
+        # Attempt to unpack major, minor, and ignore the rest
+        major, minor, *_ = version.split(".")
+    except ValueError:
+        # If unpacking fails, assume only major is provided
+        return operator, version
 
     # Return only major if minor is "0", otherwise return major.minor
     return operator, major if minor == "0" else f"{major}.{minor}"
--- a/grayskull/strategy/parse_poetry_version.py
+++ b/grayskull/strategy/parse_poetry_version.py
@@ -377,6 +377,8 @@ def combine_conda_selectors(python_selector: str, platform_selector: str):
     Combine selectors based on presence
     """
     if python_selector and platform_selector:
+        if " or " in python_selector:
+            python_selector = f"({python_selector})"
         selector = f"{python_selector} and {platform_selector}"
     elif python_selector:
         selector = f"{python_selector}"
--- a/grayskull/strategy/parse_poetry_version.py
+++ b/grayskull/strategy/parse_poetry_version.py
@@ -236,3 +236,119 @@ def encode_poetry_version(poetry_specifier: str) -> str:
         conda_clauses.append(poetry_clause)
 
     return ",".join(conda_clauses)
+
+
+def encode_poetry_platform_to_selector_item(poetry_platform: str) -> str:
+    """
+    Encodes Poetry Platform specifier as a Conda selector.
+
+    Example: "darwin" => "osx"
+    """
+
+    platform_selectors = {"windows": "win", "linux": "linux", "darwin": "osx"}
+    poetry_platform = poetry_platform.lower().strip()
+    if poetry_platform in platform_selectors:
+        return platform_selectors[poetry_platform]
+    else:  # unknown
+        return ""
+
+
+def encode_poetry_python_version_to_selector_item(poetry_specifier: str) -> str:
+    """
+    Encodes Poetry Python version specifier as a Conda selector.
+
+    Example: ">=3.8,<3.12" => "py>=38 or py<312"
+
+    # handle exact version specifiers correctly
+    >>> encode_poetry_python_version_to_selector_item("3.8")
+    "py==38"
+    >>> encode_poetry_python_version_to_selector_item("==3.8")
+    "py==38"
+    >>> encode_poetry_python_version_to_selector_item("!=3.8")
+    "py!=38"
+
+    # handle caret operator correctly
+    >>> encode_poetry_python_version_to_selector_item("^3.10")
+    # renders '>=3.10.0,<4.0.0'
+    "py>=310 or py<4"
+
+    # handle tilde operator correctly
+    >>> encode_poetry_python_version_to_selector_item("~3.10")
+    # renders '>=3.10.0,<3.11.0'
+    "py>=310 or py<311"
+    """
+
+    if not poetry_specifier:
+        return ""
+
+    version_specifier = encode_poetry_version(poetry_specifier)
+
+    conda_clauses = version_specifier.split(",")
+
+    conda_selectors = []
+    for conda_clause in conda_clauses:
+        operator, version = parse_python_version(conda_clause)
+        version_selector = version.replace(".", "")
+        conda_selectors.append(f"py{operator}{version_selector}")
+    selectors = " or ".join(conda_selectors)
+    return selectors
+
+
+def parse_python_version(selector: str):
+    """
+    Return operator and normalized version from a version selector
+
+    Examples:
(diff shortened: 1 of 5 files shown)


grayskull/strategy/parse_poetry_version.py
```diff
@@ -236,3 +236,109 @@ def encode_poetry_version(poetry_specifier: str) -> str:
...
+
+def encode_poetry_python_version_to_selector_item(poetry_specifier: str) -> str:
+    """
+    Encodes Poetry Python version specifier as a Conda selector.
+
+    Example: ">=3.8,<3.12" => "py>=38 or py<312"
+
+    # handle exact version specifiers correctly
+    >>> encode_poetry_python_version_to_selector_item("3.8")
+    "py==38"
```
marcelotrevisani, line 264: pytest is running your doctests here as well, that is why it is failing on ci

grayskull/strategy/parse_poetry_version.py
```diff
@@ -236,3 +236,109 @@ def encode_poetry_version(poetry_specifier: str) -> str:
...
+    >>> encode_poetry_python_version_to_selector_item("==3.8")
+    "py==38"
+    >>> encode_poetry_python_version_to_selector_item("!=3.8")
+    "py!=38"
+
+    # handle caret operator correctly
+    >>> encode_poetry_python_version_to_selector_item("^3.10")
+    # renders '>=3.10.0,<4.0.0'
+    "py>=310 or py<4"
```
xylar, line 273:
```suggestion
    "py>=310 and py<4"
```

grayskull/strategy/parse_poetry_version.py
```diff
@@ -236,3 +236,109 @@ def encode_poetry_version(poetry_specifier: str) -> str:
...
+
+    # handle tilde operator correctly
+    >>> encode_poetry_python_version_to_selector_item("~3.10")
+    # renders '>=3.10.0,<3.11.0'
+    "py>=310 or py<311"
```
xylar, line 278:
```suggestion
    "py>=310 and py<311"
```

grayskull/strategy/parse_poetry_version.py
```diff
@@ -236,3 +236,109 @@ def encode_poetry_version(poetry_specifier: str) -> str:
...
+    version_specifier = encode_poetry_version(poetry_specifier)
+
+    conda_clauses = version_specifier.split(",")
+
+    conda_selectors = []
+    for conda_clause in conda_clauses:
+        operator, version = parse_python_version(conda_clause)
+        version_selector = version.replace(".", "")
+        conda_selectors.append(f"py{operator}{version_selector}")
+    selectors = " or ".join(conda_selectors)
```
xylar, line 293:
```suggestion
    selectors = " and ".join(conda_selectors)
```

xylar (commented):
I believe you currently have `or` where there should be `and`, e.g. `>=3.10.0,<3.11.0` --> `py>=310 and py<311`.

xylar:
I think it may eventually be important to support `~=` but I would be happy with a solution that doesn't support that for now.  Same with `.*`.

grayskull/strategy/parse_poetry_version.py
```diff
@@ -236,3 +236,109 @@ def encode_poetry_version(poetry_specifier: str) -> str:
...
+    operator = match.group("operator")
+    # Default to "==" if no operator is provided or "="
+    operator = "==" if operator in {None, "="} else operator
+    version = match.group("version")
+
+    # Split into major, minor, and discard the rest (patch or additional parts)
+    major, minor, *_ = version.split(".")
+
+    # Return only major if minor is "0", otherwise return major.minor
+    return operator, major if minor == "0" else f"{major}.{minor}"
```
xylar, line 327: I think we probably need this to also support:
```
parse_python_version('>=3')
```
When I tried that, I see (as I expected I might):
```
  File "/home/xylar/code/grayskull/poetry-deps-w-python-ver-platform-and-multiple-deps/./test.py", line 6, in <module>
    print(parse_python_version('>=3'))
          ~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/home/xylar/code/grayskull/poetry-deps-w-python-ver-platform-and-multiple-deps/grayskull/strategy/parse_poetry_version.py", line 324, in parse_python_version
    major, minor, *_ = version.split(".")
    ^^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected at least 2, got 1)
```
Could you add a test for this situation and make the necessary changes to support a single digit as well?

xylar:
@lorepirri, I should have said before diving into the details.  I really appreciate this work a ton!  The lack for support for these python selectors has been bothering me a lot for a long time, but I've felt a little overwhelmed when I looked into trying to fix them.  This is some great work!  Not surprisingly, there are still a few wrinkles to iron out.

grayskull/strategy/parse_poetry_version.py
lorepirri: Added support for 1 digit and three dots case (e.g. `3.8.0.1`, never seen for Python version specifier but valid syntax). Thanks for checking that out!

grayskull/strategy/parse_poetry_version.py
lorepirri: fixed

grayskull/strategy/parse_poetry_version.py
lorepirri: fixed

grayskull/strategy/parse_poetry_version.py
lorepirri: fixed

lorepirri:
> I believe you currently have `or` where there should be `and`, e.g. `>=3.10.0,<3.11.0` --> `py>=310 and py<311`.

oh yeah... sorry about that! Nice catch. The `or` got stuck in my head... I worked the whole time with `py<37 or py>=310` in mind (mostly coming from the `skip: True`), then did not test that case. 

I'm left wondering how Poetry will interpret `<3.7,>=3.10` 🤔 is every comma an `and`, or do they check whether the lower/upper bounds converge? Probably.

lorepirri:
Added support for `or`. Updated the description of the PR.

Multiple version specifiers are at the moment always rendered in `and`. For example `>=3.10.0,<3.11.0` with render `py>=310 and py<311`. Same as `conda-build` is doing. Ref: https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications

If one would want to express `<3.7` or `>=3.10` should use the pipe operator `<3.7|>=3.10`.

`<3.7|>=3.10` renders `py<37 or py>=310` and `<3.8|>=3.10,!=3.11` renders `py<38 or py>=310 and py!=311` (`,` and `and` take precedence over `|` and `or`).

marcelotrevisani:
> When there are any extras in a dependency requirement for `pypi` the `config.is_arch` is set to `True`: https://github.com/conda/grayskull/blob/main/grayskull/strategy/pypi.py#L355
> 
> I did not explore the reason yet.
> 
> Shall it be the same for Poetry? Maybe this is another question @marcelotrevisani can help with.

that is set to ``True`` because when that function is called it means selectors are in use, and if we have selectors the recipe cannot be no arch. However, a few things have changed in conda-forge and that is not entirely true anymore. We allow a few selectors in some scenarios, but as a rule of thumb, if you don't want to dive into it, that is fine to still consider to be arch if it has selectors.

And thanks a lot for your contribution! I really appreciate it, I believe @xylar have done a good review and I am also happy with the current state of the PR, unless you want to improve a few bits more, but that can also be done in a second PR if you wish :)

xylar:
I imagine there could be some growing pains with this update.  There are a lot of cases to consider and @lorepirri, you've done quite a good job of that. Even so, if there are some bugs that come up after we send this into the wild, I'm more than happy to help with them in the future.

xylar:
@lorepirri, it sounded like you wanted to tackle `~=` and I think that would be great.  Once that's in, I'm happy to approve.

lorepirri:
> add "~=" and ".*" support

@xylar I've [added "~=" and ".*" support](https://github.com/conda/grayskull/pull/592/commits/ee3a6d6a85fdcb0d786360bb2f26b776d54ee34a) it was a bit more effort than I thought mostly because ".*" expands to two selectors in "or". The PR starts to be a bit crowded, I would not add anything else here. There are a few changes to review, unfortunately.

> I imagine there could be some growing pains with this update. There are a lot of cases to consider and @lorepirri, you've done quite a good job of that. Even so, if there are some bugs that come up after we send this into the wild, I'm more than happy to help with them in the future.

Thanks for your support! I made the parser a little bit more robust and now it accepts a PEP440 version in the Python version even though I would not expect to ever receive `2!3.12.2.post1` in it (it will work and render `py==312`, though).

> > When there are any extras in a dependency requirement for `pypi` the `config.is_arch` is set to `True`: https://github.com/conda/grayskull/blob/main/grayskull/strategy/pypi.py#L355
> > I did not explore the reason yet.
> > Shall it be the same for Poetry? Maybe this is another question @marcelotrevisani can help with.
> 
> that is set to `True` because when that function is called it means selectors are in use, and if we have selectors the recipe cannot be no arch. However, a few things have changed in conda-forge and that is not entirely true anymore. We allow a few selectors in some scenarios, but as a rule of thumb, if you don't want to dive into it, that is fine to still consider to be arch if it has selectors.
> 
> And thanks a lot for your contribution! I really appreciate it, I believe @xylar have done a good review and I am also happy with the current state of the PR, unless you want to improve a few bits more, but that can also be done in a second PR if you wish :)

Thanks for having created `grayskull` in the first place and thank you for the support! I've changed a few things to add "~=" and ".*" support. I hope the code did not become cumbersome.

As for `config.is_arch` is set to `True`, let me see if I can quickly bring in `config` because it's not there yet.

xylar (approved):
@lorepirri, I'm very happy with the work you have done on the `~=` and `.*` support.  It's clear that some choices need to be made about how these get supported in a way that makes sense for python and I think the choices you've documented make a lot of sense.  Hopefully, no one will be silly enough to use some of the weirder cases you added support for but it's nice to have some sensible fallback even if they do rather than just having grayskull crash.

lorepirri:
PR description updated (removed the TODOs and ~strikethrough~ leftovers).

> It does handle compatible release operator `~=` correctly and the wildcard `.*` suffix in the version as well (not very common though as the resulting selectors are not interesting).

> 👉 Fixes a minor bug: the compatible release operator `~=` in the dependency version could not be used because the parser was jumping to expanding the tilde `~` and crashing on `=` (or returning the wrong version containing `=`). Therefore I suspect that the  `~=` operator was never handled (maybe never used, or nobody reported this bug).

lorepirri:
> @lorepirri, I'm very happy with the work you have done on the `~=` and `.*` support. It's clear that some choices need to be made about how these get supported in a way that makes sense for python and I think the choices you've documented make a lot of sense. Hopefully, no one will be silly enough to use some of the weirder cases you added support for but it's nice to have some sensible fallback even if they do rather than just having grayskull crash.

Thank you @xylar for having had a look at it once again! I'm glad that I could add some value to the project. I decided to simplify the cases for the `~=` and `.*` anticipating how those were actually rendered.

In case there will be the need to fully expand `.*` for example for the dependency versions, one could use this function to get the full left and right bounds (I leave it here because I wrote it but it was not useful at the end, other than testing my theory):

```
def wildcard_version_bounds(wildcard_version: str) -> tuple[str | None, str | None]:
    """

    >>> wildcard_version_bounds("*")
    (None, None)

    >>> wildcard_version_bounds("1.*")
    ('1.0a0', '2')

    >>> wildcard_version_bounds("1.1.*")
    ('1.1.0a0', '1.2')

    >>> wildcard_version_bounds("1.1.1.*")
    ('1.1.1.0a0', '1.1.2')

    # Python versions
    >>> wildcard_version_bounds("3.*")
    ('3.0a0', '4')

    >>> wildcard_version_bounds("3.12.*")
    ('3.12.0a0', '3.13')

    >>> wildcard_version_bounds("3.9.1.*")
    ('3.9.1.0a0', '3.9.2')

    """
    if "*" not in wildcard_version:
        raise ValueError("The version must contain a wildcard '*'.")

    # Remove the wildcard and parse the base version
    base = wildcard_version.rstrip(".*")
    lower_bound = f"{base}.0a0" if base else None  # Lowest possible version

    if not base:
        upper_bound = None  # No restriction for `*`
    else:
        # Parse base as a Version object and increment the last meaningful part
        base_version = Version(base)
        parts = list(base_version.release)
        parts[-1] += 1  # Increment the last meaningful part
        # Construct the upper bound by resetting all lower components to 0
        upper_bound = ".".join(map(str, parts))

    return lower_bound, upper_bound
```

grayskull/strategy/parse_poetry_version.py
```diff
@@ -236,3 +247,480 @@ def encode_poetry_version(poetry_specifier: str) -> str:
...
+    encode_poetry_version is because the selectors for the python
+    version use only major and minor, and therefore the compatible
+    release operator ~= makes sense only in the case of major and
+    minor specified (e.g. "~=3.8" -> "py>=38 and py<4") and just by
+    knowing that we can avoid having to detect cases like:
+        "~=3.8.0.1" -> ">=3.8.0.1, ==3.8.0.*" -> ">=3.8.0.1, <3.8.1.0a" -> "py==38"
+    and expand only for:
+        "~=3.8" -> ">=3.8, ==3.*" -> ">=3.8, <4.0a" -> "py>=38 and py<4"
+        "~=3.0" -> ">=3.0, ==3.*" -> ">=3.0, <4.0a" -> "py>=3 and py<4"
+    in the rest of the cases it's a simple conversion to ">=" operator.
```
lorepirri, line 467:
```suggestion
    in the rest of the cases it's a simple conversion to ">="/"==" operator.
```

marcelotrevisani (approved):
LGTM! Thanks for your contribution!

lorepirri:
Thank you for merging @marcelotrevisani . We could close #463 #535 and the PR #523 (it seems stale).

PR #597, opened 2024-12-29
Fix support for `pyproject.toml` dependencies with multiple selectors

xylar:
### Description

It is common for the same dependency to have multiple selectors, like the following:
```
    - pandas >=1.5.3,<2.2  # [py<39]
    - pandas >=2.1.2,<2.2  # [py>=39]
```
This merge attempts to fully support this capability.

In several places in the code where dependencies are merged, care is now taken to allow multiple versions of the same package as long as they come from the same source.  This happens when merging `toml` and `setup` dependencies into `sdist` dependencies, and again when merging `sdist` dependencies with `pypi` dependencies.  It also takes place when removing redundancies later in the process of building a recipe.

In addition, because some strategies (namely parsing of `poetry` dependencies) add selectors early in the process, care has been taken not to drop selectors that may mistakenly be identified as tags or comments.

Finally, a fix has been added to identify that a package `is_arch` when it has selectors, just as it is identified as `is_arch` when "extras" are detected.

Four new pypi tests have been added and new testing has been added to the existing `test_run_requirements_sdist` test to ensure that multiple selectors for the same package are handled correctly.

### DIFF
--- a/grayskull/strategy/py_base.py
+++ b/grayskull/strategy/py_base.py
@@ -709,10 +709,12 @@ def merge_deps_toml_setup(setup_deps: list, toml_deps: list) -> list:
     for dep_name, dep in zip(setup_dep_names, setup_deps):
         if not dep_name.strip():
             continue
-        alternatives = [dep_name, dep_name.replace("_", "-"),
-                        dep_name.replace("-", "_")]
-        found = any([alternative in toml_dep_names for alternative in
-                     alternatives])
+        alternatives = [
+            dep_name,
+            dep_name.replace("_", "-"),
+            dep_name.replace("-", "_"),
+        ]
+        found = any([alternative in toml_dep_names for alternative in alternatives])
         # only add the setup dep if no alternative name was found
         if not found:
             merged_deps.append(dep)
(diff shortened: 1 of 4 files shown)


xylar:
@marcelotrevisani, as always, I would very much appreciate your input and review.

@lorepirri, this work builds on yours in #592.  I would very much appreciate your feedback as well.

I have put several days into learning my way around `grayskull` (and `souschef`), but I have to return to my normal work soon so I don't know how much time I'll be able to put into this in the next while.  That's just to say that, if I'm slow to respond, don't take that as a lack of interest.

marcelotrevisani (approved):
LGTM! Thanks again! :)

marcelotrevisani:
> @marcelotrevisani, as always, I would very much appreciate your input and review.
> 
> @lorepirri, this work builds on yours in #592. I would very much appreciate your feedback as well.
> 
> I have put several days into learning my way around `grayskull` (and `souschef`), but I have to return to my normal work soon so I don't know how much time I'll be able to put into this in the next while. That's just to say that, if I'm slow to respond, don't take that as a lack of interest.

the modification looks sensible to me, I believe we can integrate it, thanks!

Don't worry, I am not taking that as a lack of interest at all, I am very glad to the insights and inputs you are bringing to gray skull!

xylar:
@marcelotrevisani, wonderful, thanks so much.  I'd like to follow up at some point when I have time with some updates to `--strict-conda-forge` but I think that's also a project for when I have a few free days.

PR #598, opened 2025-01-05
Remove one layer of indirection in pypi generated recipes

hmaarrfk:
### Description

`{{ name[0] }}/{{ name }}` is uncessary and adds to cognitive load.

Thanks for considering.


marcelotrevisani:
Hi @hmaarrfk , can you fix the failing tests please?

cc: @ocefpaf

ocefpaf:
@hmaarrfk I'm +1 to that change.  Some folks where against it when we first tried to implement that a long time ago. Maybe, as we shift to v1 recipes, we can try again and see how it goes.

PR #605, opened 2025-02-17
Treat SSL errors as not available

jamesmyatt:
SSL errors might occur due to firewall issues. So just assume it's not available, since it only drives the CLI output.

It would be better to be able to check against a different repository, such as prefix.dev, but they aren't guaranteed to use the same format URLs when just checking whether there are any packages with the same name.


marcelotrevisani:
Hi, sorry for my delay, it looks sensible to me.

if all tests passes I will merge it, thanks! :)

marcelotrevisani (approved):

PR #606, opened 2025-03-15
add custom index arguments

jamesfricker:
### Description

I'd like to be able to generate conda recipes inside my company. I'm not able to do this at the moment, since `anaconda.org` is hard-coded into the application and we do not have external internet access on dev machines. 

This PR allows users to pass a set of custom channels to search for existing packages.


marcelotrevisani (approved):
LGTM! Thanks for your contribution!

PR #627, opened 2025-10-04
fix(track_packages): remove deprecated pkg_resources usage

nachofest:
### Description

Replace deprecated `pkg_resources.parse_version` with `packaging.version.Version` in `grayskull/base/track_packages.py` to remove the setuptools deprecation warning.

**What changed**
- `from pkg_resources import parse_version` → `from packaging.version import Version`
- Updated version comparisons to use `Version(...)` 

**Why**
- `pkg_resources` will be deprecated;

### DIFF
--- a/grayskull/base/track_packages.py
+++ b/grayskull/base/track_packages.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass
 from functools import lru_cache
 from pathlib import Path
 
-from pkg_resources import parse_version  # noqa
+from packaging.version import Version  # noqa
 from ruamel.yaml import YAML
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


jdnewmil:
Just to bump... the time window for the pkg_resources deprecation is expected to close in less than six weeks after this bump:

> UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.

marcelotrevisani:
Thank you for your contribution, sorry for my huge delay, I am a bit busy and with other things on my plate at the moment.
Thank you again!

PR #653, opened 2026-05-02
fix: update URL for fetching open source license data

beckermr:
### Description

The OSI folks changed their API endpoint URL and the data it returns. See this blog post: https://opensource.org/blog/introducing-the-new-api-for-osi-approved-licenses. 

It appears the old data had more information on alternative license names compared to the new data. 

I added the missing other license names back to the code.


maresb:
Works for me, thanks for fixing this @beckermr!

Original exception:

```
      raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)                                                                                                                                           
  requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)  
```

beckermr:
We can comment on this forum to ask them about the old data: https://discuss.opensource.org/t/introducing-the-new-api-for-osi-approved-licenses/1169/4

I cannot figure out how to join however, so if someone else has an account already, they should go ahead!

maresb:
What do you mean "ask them about the old data"?

beckermr:
The old API data had an "other_names" column that had historical names used by licenses. For example, MIT is called Expat by some folks. This information is now gone in the new API.

beckermr:
See eg this bit of code: https://github.com/conda/grayskull/blob/eb58ea48f43fa43630373fe6fea7af2910f84c14/grayskull/license/licence_cache.json#L1599

beckermr:
We can use the json blob we have to build a mapping to the other names, but if we are missing it then maybe others are too?

beckermr:
Well actually there are only a few entries that have other names. I will add it back by hand to this PR.

beckermr:
I have to say though, they removed a lot of useful information like alternative identifiers when they changed the API. :/

tests/license/test_discovery.py
```diff
@@ -46,7 +46,7 @@ def spdx_org_license_mit() -> list:
 
 def test_match_license():
     assert match_license("MIT License")["licenseId"] == "MIT"
-    assert match_license("Expat")["licenseId"] == "MIT"
+    assert match_license("The MIT License")["licenseId"] == "MIT"
```
beckermr:
```suggestion
    assert match_license("Expat")["licenseId"] == "MIT"
```

tests/license/test_discovery.py
```diff
@@ -65,8 +65,8 @@ def test_get_opensource_license_data():
 @pytest.mark.parametrize(
     "licence_name, short_licence",
     [
-        ("MIT License", "MIT"),
         ("Expat", "MIT"),
+        ("The MIT License", "MIT"),
```
beckermr:
```suggestion
        ("MIT License", "MIT"),
```

tests/license/test_discovery.py
```diff
@@ -65,8 +65,8 @@ def test_get_opensource_license_data():
 @pytest.mark.parametrize(
     "licence_name, short_licence",
     [
-        ("MIT License", "MIT"),
         ("Expat", "MIT"),
+        ("MIT License", "MIT"),
```
beckermr:
```suggestion
        ("MIT License", "MIT"),
        ("Expat", "MIT"),
```

beckermr:
Alright this is ready to merge @conda/grayskull. I have kept the ability to parse the older data format so that if the old data is ever added back, we can use. I have put the info from the old data into a dict in the python so that we still have it.

marcelotrevisani (approved):

Issue #383, opened 2022-08-31
[BUG] metadata package name wrong when the recipe has multuiple outputs


woutdenolf:
For example

```
grayskull pypi pypushflow --extras-require-all --extras-require-split --extras-require-test test --extras-require-exclude dev --strict-conda-forge
```

gives this

```yaml
{% set name = "pypushflow" %}

package:
  name: {{ name|lower }}
  version: {{ version }}
```

but it should be this

```yaml
{% set name = "pypushflow" %}

package:
  name: {{ name|lower }}-meta  # !!!!
  version: {{ version }}
```

The reason is this part of the code (it only uses the version, not the name):

```
def update_recipe(recipe: Recipe, config: Configuration, all_sections: List[str]):
            if section == "package":
                set_global_jinja_var(recipe, "version", metadata["package"]["version"])
                config.version = metadata["package"]["version"]
                recipe["package"]["version"] = "<{ version }}"
```

PR #654, opened 2026-05-05
Fix sdist package name

beenje:
### Description

In https://github.com/marcelotrevisani/souschef/pull/64, the `name` variable was removed from the generated recipe.
This broke recipe generation from local sdist when the package name and sdist filename didn't match.

`grayskull` was still inserting the correct name as variable but it wasn't used anymore:

```
{% set version = "1.9.1.dev1+gfbc6c9ac2" %}
{% set name = "sardana-flexpes" %}

package:
  name: sardana_flexpes
  version: {{ version }}
```

Instead of using `set_global_jinja_var(recipe, "name",...)`, we now update directly the `name` key in the `package` section.

While fixing this bug, I saw another issue:

```
 package_metadata["name"] = package_metadata["name"].replace(
     config.name, "<{ name|lower }}"
 )
```

This would use  the `name` variable, which isn't defined anymore, leading to a broken recipe.
I couldn't reproduce the initial issue https://github.com/conda/grayskull/issues/383. I assume that the `package_metadata["name"]` is already what we want and we don't need to update anything?

@woutdenolf do you have any input on this?

### DIFF
--- a/grayskull/strategy/pypi.py
+++ b/grayskull/strategy/pypi.py
@@ -588,16 +588,11 @@ def update_recipe(recipe: Recipe, config: Configuration, all_sections: Iterable[
         if metadata.get(section):
             if section == "package":
                 package_metadata = dict(metadata[section])
-                if package_metadata["name"].lower() == config.name.lower():
-                    # In case of from_local_sdist, the initial name set in the recipe
-                    # came from the sdist filename and shall be updated.
-                    # No need otherwise.
-                    if not config.from_local_sdist:
-                        package_metadata.pop("name")
-                else:
-                    package_metadata["name"] = package_metadata["name"].replace(
-                        config.name, "<{ name|lower }}"
-                    )
+                # In case of from_local_sdist, the initial name set in the recipe
+                # came from the sdist filename and shall be updated.
+                # No need otherwise.
+                if not config.from_local_sdist:
+                    package_metadata.pop("name")
 
                 set_global_jinja_var(recipe, "version", package_metadata["version"])
                 config.version = package_metadata["version"]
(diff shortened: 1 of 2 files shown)


woutdenolf:
Oef too long ago sorry. Does `package_metadata` come from `conda-souschef` and `package_metadata["name"]` no longer exists? Is that it? If yes, why do we still need `package_metadata.pop("name")`?

beenje:
> Oef too long ago sorry. Does `package_metadata` come from `conda-souschef` and `package_metadata["name"]` no longer exists? Is that it? If yes, why do we still need `package_metadata.pop("name")`?

No, `package_metadata` comes from `get_metadata(recipe, config)`. So `name` should be correct (metadata is merged from different sources).
That key has already been set when creating the recipe with `conda-souschef` and should match in most cases.
For local sdist it has to be updated but isn't needed otherwise. So I kept the pop to minimise the changes.
I still don't really understand when we could get `package_metadata["name"].lower() != config.name.lower()`.

But I realise that we had the `pop` before because name was set to the Jinja expression `{{ name }}`, which isn't the case anymore.
So updating  everytime should be fine. I'll remove it.

woutdenolf:
> I still don't really understand when we could get package_metadata["name"].lower() != config.name.lower().

Ah yes, it used to be the case that the conda package name was not necessarily equal to the package name from pypi, sdist or wherever it came from.

See https://github.com/conda/grayskull/issues/388: in case of  `grayskull pypi ... --extras-require-split` and if there are several "outputs" (so several conda packages, hence the "split") then the name of the main package returned by  `get_metadata` used to be "{name}-meta".

The "-meta" suffix got removed in https://github.com/conda/grayskull/commit/389647a741a8950f3f9bd56208884c94fb6eda43.

So this PR should be ok.

